PR #562, opened 2019-05-28
mbl-signing: Add a Python wrapper for the ATF fiptool.

rwalton-arm:
Wrapper for the ATF fiptool, used for managing FIP image blobs.

Adds the initial code and folder structure for "mbl-signing-lib";
a Python library to be used for code signing.

fip/fiptool.py: Add a fiptool module to wrap ATF fiptool functionality.
conf/toolpaths.py: Paths to the tools mbl-signing-lib uses/will use.
fip/test_fiptool.py: Tests for the fiptool module.

JIRA: IOTMBL-1968


openembedded-core-mbl/scripts/lib/mbl-signing/signing/fip/fiptool.py
```diff
@@ -0,0 +1,389 @@
...
+    output_fip_filename: pathlib.Path,
+    align: int = 0,
+    blob_uuid: str = "",
+    blob_file: str = "",
+    plat_toc_flags: int = 0,
+) -> int:
+    """Create a new FIP with the given images.
+
+    :param ImageSpec img_spec: ImageSpec object.
+    :param str output_fip_filename: Name to give the output FIP image.
```
diegorusso: What about the other parameters?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/fip/fiptool.py
```diff
@@ -0,0 +1,389 @@
...
+
+    This dict-like object is used to invoke the ATF fiptool with the
+    correct options.
+
+    When this object's str method is called it produces an option string
+    intended to be passed to the fiptool CLI.
+
+    __setitem__ has been overridden to check for 'valid opts' i.e valid keys.
+
+    This object is intended to be subclassed, the valid_opts attribute should
```
diegorusso: maybe we need to define an __init__ which raises a NotImplementedError exception

openembedded-core-mbl/scripts/lib/mbl-signing/signing/fip/fiptool.py
```diff
@@ -0,0 +1,389 @@
...
+        msg = (
+            "{attr} is not in the known valid options.\n"
+            "Known options are:\n {valid_opts}".format(
+                attr=attribute, valid_opts="\n".join(opt_dict.valid_opts)
+            )
+        )
+        super().__init__(msg)
+
+
+# Validate the fiptool exists and can be invoked at module import time.
```
diegorusso: is it not better to wrap it in a function and call that function?

rwalton-arm:
> How do we use this library?

The library will be used in various bbclasses in place of the current invocations of the fiptool; I'll be raising PRs related to that in the next sprint. It will also be used in the standalone image signing tool I'll be working on in the sprint after next.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/fip/fiptool.py
rwalton-arm: Instead I've marked the `__str__` method as an `@abstractmethod' which prevents the class being instantiated by raising a `TypeError` (and any attempted instantiation will be flagged in linters such as flake8/pylint).

PR #595, opened 2019-06-24
Temporary fix for cryptsetup to build natively

jjohnson-arm:
Do not include udev in the PACKAGECONFIG as it can't be built natively.
Temporary workaround so that we don't have to pin meta-openembedded.
Future work will be done to fix/change dm-verity operation.

IOTMBL-2206


jjohnson-arm:
This should work as it is currently building: http://jenkins.mbed-linux.arm.com/blue/organizations/jenkins/mbl-master/detail/mbl-master/1353/pipeline/
And it is a similar fix to the bbappend I did before the weekend which worked [fine](http://jenkins.mbed-linux.arm.com/blue/organizations/jenkins/mbl-master/detail/mbl-master/1351/pipeline)

PR #606, opened 2019-07-10
optee-os: Fix imx8 failure after reboot many times

niej:
Update optee-os to fix imx8 failure after reboot many times.

Memory content from previous boot should be cleaned so that dtb overlay will not be added to top of previous dtb overlay again and again.


niej:
test report:
http://e120856-lin.cambridge.arm.com/results.php?image_version=custom&image_number=4&custom_version=jun-imx8-reboot-failure&lava_query=&lava_owner=&failures=false&submitter=junnie01&sort=by_job

meta-linaro-mbl/meta-optee/recipes-security/optee/optee-os/0002-core-clear-original-dtb-data-in-special-case.patch
```diff
@@ -0,0 +1,37 @@
...
+--- a/core/arch/arm/kernel/generic_entry_a64.S
++++ b/core/arch/arm/kernel/generic_entry_a64.S
+@@ -56,6 +56,10 @@ FUNC _start , :
+ 	mov	x19, x0		/* Save pagable part address */
+ #if defined(CFG_DT_ADDR)
+ 	ldr     x20, =CFG_DT_ADDR
++#ifndef CFG_DT_OVERLAY
++	mov	x14, #0x0       /* clear dtb magic value */
++	str	x14, [x20]
++#endif
```
bryanodonoghue: This confuses me.
Aren't we using DTB overlay on i.MX8 ?

niej: We are using DTB overlay with CFG_EXTERNAL_DTB_OVERLAY, but not CFG_DT_OVERLAY.

grandpaul: I also got questions to this. As you modified a64 code. But we also got the same problems on arm-32 platforms. How do we address it?

niej: Because there is no suitable configuration to limit the code and the logic is politics as you mention. Also because this PR is a fix to corner case of test system, so I assume this is a temp solution to imx8mm.
Anyway, I already update patch with change in init_overlay code, although the condition on CFG_DT_ADDR still breaks current optee document.

bryanodonoghue: Lets just put the patch into our Linaro tree.

bryanodonoghue: That way we can go back and look at this again if we have to, or propose it for upstream.

grandpaul: ok. Agree. Let's just keep this patch locally.

niej:
jenkins build and lava test:

http://jenkins.mbed-linux.arm.com/view/JunNie/job/jun-imx8-reboot-failure/5/

http://e120856-lin.cambridge.arm.com/results.php?image_version=custom&image_number=5&custom_version=jun-imx8-reboot-failure&lava_query=&lava_owner=&failures=false&submitter=junnie01&sort=by_job

grandpaul:
I do test the jenkins build jun-imx8-reboot-failure/7/.
It does boot over 100 times.
I'm using the following scripts.


/home/root/reboot_test.sh
```
#!/bin/sh

CONFIG_FILE=/home/root/reboot_test.conf

if [[ -f $CONFIG_FILE ]]; then
    . $CONFIG_FILE
else
    COUNTER=0
fi


COUNTER=`expr $COUNTER + 1`
echo "COUNTER="$COUNTER > $CONFIG_FILE

sleep 60
shutdown -r 1
sleep 1000
```

/lib/systemd/system/reboot_test.service
```
[Unit]
Description=Reboot test

[Service]
ExecStart=/home/root/reboot_test.sh

[Install]
WantedBy=multi-user.target
```

bryanodonoghue:
Removes the previous bug I could trigger with holding down return on the keyboard somehow so merged.

PR #617, opened 2019-07-23
Adjust partition variables to allow computation of offsets

jonathanhaigh-arm:
We want to be able to change partition offsets based on the flash erase
block size. In order to do that we need to be able to calculate e.g. the
BL3/WKS_BOOTLOADER2 offsets rather than rely on the MACHINE config
setting them.

In order to do this, we need to differentiate between bootloader
alignments and bootloader offsets, so introduce variables for offsets.

For consistency, add extra variables for all of our non-bootloader
partitions so that they can be adjusted in the same ways as our
bootloader partitions.

The logic for dealing with these variables is a lot more complicated
now, so put it in Python functions and generate a dictionary for each
partition containing all the info.

To keep all of our partition info in one place, move the info from fstab
for each partition into the mbl-partitions class and generate our fstab
lines from the dicts we now have for each partition.

Add flash erase block size info for our platforms that have eMMC.

For IOTMBL-2280: Update .wks files for new partition layout guidelines

**Jenkins Links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/122/ (Aborted - hit [IOTMBL-2334](https://jira.arm.com/browse/IOTMBL-2334) on imx8).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/123/ (Succeeded).

**Testing**
* Lava results: http://e120856-lin/results.php?image_version=custom&image_number=123&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. There were a bunch of failures - the usual Wi-Fi test failures and timeouts after writing to flash. The timeouts could be related to my changes here - to investigate.
* Second run: http://e120856-lin/results.php?image_version=custom&image_number=124&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. More failures - flakiness when flashing the image on rpi3 and imx8 and SD cards in imx8 that are too small. https://github.com/ARMmbed/mbl-tools/pull/265 Should help with the general flakiness and @ollivainola is putting bigger SD cards in the imx8s.
* Third run: http://e120856-lin/results.php?image_version=custom&image_number=125&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. Some more failures here, but all signs are pointing to infrastructure issues.


meta-mbl-distro/recipes-core/packagegroups/packagegroup-mbl-development.bb
```diff
@@ -26,6 +26,7 @@ PACKAGEGROUP_MBL_DEVELOPMENT_PKGS_append = " gobject-introspection"
 PACKAGEGROUP_MBL_DEVELOPMENT_PKGS_append = " e2fsprogs"
 PACKAGEGROUP_MBL_DEVELOPMENT_PKGS_append = " mbed-crypto-test"
 PACKAGEGROUP_MBL_DEVELOPMENT_PKGS_append = " memtester"
+PACKAGEGROUP_MBL_DEVELOPMENT_PKGS_append = " mmc-utils"
```
diego-sueiro: Why do we need this package on the rootfs?

jonathanhaigh-arm: We don't. I've just got that there temporarily to check the flash erase block sizes on each platform (that has eMMC).

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+# Partition variables
+# ------------------------------------------------------------------------------
 # We have two different flavours of "partition" in MBL:
 # * Normal file system partitions, recorded in the partition table.
 # * Raw non-file system "partitions", not recorded in the partition table.
+# 
+# List of all partitions, in order
+MBL_PARTITION_NAMES = ""
+#
+# For each of these "partitions" there are the following variables:
```
diego-sueiro: Which "partitions"?
The line above is setting it to empty.
Maybe we need a better comment about what are the default/expected/supported partition layout. Or a specific documentation file to give a better background.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+# * MBL_<partition>_ALIGN_KiB   - minimum alignment in KiB
+# * MBL_<partition>_SIZE_KiB    - size in KiB
+# * MBL_<partition>_SIZE_MiB    - size in MiB
+# * MBL_<partition>_OFFSET1_KiB - offset of first bank in KiB
+# * MBL_SKIP_<partition>        - skip this partition?
```
diego-sueiro: Can we follow the same pattern of the other variables, where we have `MBL_<partition>_*`?
I think we should remove the question mark.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+# * MBL_<partition>_NO_FS       - is this a raw partition (not a file system partition)?
```
diego-sueiro: I think we should remove the question mark.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+# * MBL_<partition>_FS_FREQ     - fifth fstab(5) field
+# * MBL_<partition>_FS_PASSNO   - sixth fstab(5) field
+#
+# Notes:
+# * MBL_<partition>_ALIGN_KiB may be left unset outside of this file and a
+#   default value will be used. For WKS_BOOTLOADER1 the default value is "1".
+#   For other partitions the default value is ${MBL_FLASH_ERASE_BLOCK_SIZE_KiB}.
+#
+# * If both MBL_<partition>_SIZE_KiB and MBL_<partition>_SIZE_MiB are set
+#   then they must not contradict each other.
```
diego-sueiro: What is the meaning of contradicting? They can be different?
Why we don't reduce implementation complexity and just accept the size in KiB?

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I've partially addressed this with https://github.com/ARMmbed/meta-mbl/pull/617/commits/80acca9bf0501943da70a460f557e8d80dca454d.

There will be documentation (in the docs repo) about the partition layout and how it can be adjusted (https://jira.arm.com/browse/IOTMBL-531).

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/617/commits/d70b2d4558985d027806009014b90261f4b894d8

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/617/commits/d70b2d4558985d027806009014b90261f4b894d8

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+#
+# * If both MBL_<partition>_SIZE_KiB and MBL_<partition>_SIZE_MiB are left
+#   unset outside of this file then default sizes will be used.
+#
+# * For file system partitions, the size must be a multiple of 1MiB.
+#
+# * If WKS_BOOTLOADER1 is not skipped then WKS_BOOTLOADER1_OFFSET1_KiB
+#   must be set outside of this file - there is no default value because this
+#   is always board specific. For WKS_BOOTLOADER1 and WKS_BOOTLOADER_FS,
```
diego-sueiro: Did you mean `WKS_BOOTLOADER2`?

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: It's supposed to mean: if you specify the size in KiB and the size in MiB then
```
MBL_<partition>_SIZE_KiB == 1024 * MBL_<partition>_SIZE_MiB
```
must be true.

Yes, we could just accept the size in KiB. I was keeping `_SIZE_MiB` because:
* People will expect to be able to set the sizes in MiB because that's what Wic does.
* Most of the time, the size _must_ be a multiple of 1MiB (because that's what Wic uses).

It's just the bootloaders in non-file system partitions that are different.

If you think it would be clearer to just drop the `SIZE_MiB` variables then I'll gladly do that - it would make the code a lot cleaner.

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: Yes. https://github.com/ARMmbed/meta-mbl/pull/617/commits/69b0c1c8779d30746ad02fb530710d8b60023428.

meta-mbl-distro/classes/mbl-partitions.bbclass
diego-sueiro: IDK.
I like the idea of keeping it simple, but at the same time, we will break the compatibility with the standard wic parameters.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,405 @@ inherit mbl-artifact-names
...
+                bb.fatal("{} (\"{}\") must contain only decimal digits".format(var, var_str))
+                return 0
+            return int(var_str)
+
+        def _is_var_set(self, var):
+            return bool(self.d.getVar(var, True))
+
+        def _get_bool_var(self, var):
+            val = self.d.getVar(var, True)
+            return val is not None and bool(int(val))
```
rwalton-arm: `bool(int(val))` could do with some explanation as it looks a bit strange to me. Should probably add explanatory docstrings to each method of this class. Some comments in the larger methods describing what is going on would also be useful.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,405 @@ inherit mbl-artifact-names
...
-#
-# Each of these partitions can be removed from the .wks file by setting the
-# MBL_WKS_SKIP_BOOTLOADERx variable to "1" which causes an
-# MBL_WKS_COMMENT_IF_SKIP_BOOTLOADERx to be set to be the WKS comment leader
-# "#" so that the line for that partition gets commented out of the .wks file.
-#
-# Each of these partitions also has an "MBL_WKS_BOOTLOADERx_ALIGN_KiB" variable
-# to specify where it should live in flash.
+python __anonymous() {
+    class MaybeBankedPartition:
```
rwalton-arm: Why not just name this class `Partition`? `MaybeBankedPartition` is a bit clunky.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,405 @@ inherit mbl-artifact-names
...
+                self.align_KiB,
+                self.size_KiB,
+                ["{}KiB".format(offset) for offset in self.offsets_KiB]
+            )
+
+    part_names = d.getVar("MBL_PARTITION_NAMES", True).split()
+    parts = []
+    prev_part = None
+    for part_name in part_names:
+        part = MaybeBankedPartition(d, part_name, prev_part)
```
rwalton-arm: This seems strange; you pass an instance of `MaybeBankedPartition` in as an argument to another instance of `MaybeBankedPartition` so the object has knowledge of the previous partition size and offset right? Does it make sense to have another `Partitions` class which contains a list of `MaybeBankedPartition` classes, and have the `Partitions` parent class do the alignment and offset fiddling? I think the approach you've chosen here makes the code harder to understand in general.

rwalton-arm:
Sorry I'm not @diego-sueiro but I was reading through this and had a couple of comments on the Python code, so I have left a review anyway. If this is all going to change just dismiss the review and I'll re-review when the "PR ready" code is posted.

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: > Should probably add explanatory docstrings to each method of this class

Yep, that's on my list of things to do.

> `bool(int(val))`

Thanks, I'll make sure I explain this (or change it if I can no longer justify it to myself!).

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: Mmm, Yeah, I don't like the name. It's called that because an object might actually represent a pair of partitions rather than a single one, so I didn't just want to call it `Partition`.

I haven't been consistent about this distinction though... `MBL_PARTITION_NAMES` only contains a single name for each pair of banked partitions and `MBL_PARTITIONS` only contains a single entry for each pair of banked partitions. I think it would complicate the implementation too much to have separate entries for each bank though.

Maybe I should just call it `Partition` and do a better job of documentation... If you have any ideas I'll be glad to hear them!

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I sort of agree... I think the class is conflating several things:
1. Validation of BitBake variable values.
2. Setting/calculating BitBake variable values when they haven't been set explicitly by e.g. a MACHINE config.
3. Providing an interface to access partition information.

I'll see if I can come up with something better...

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -24,138 +24,407 @@ inherit mbl-artifact-names
...
+MBL_WKS_BOOTLOADER_FS_LABEL ?= "bootloaderfs"
+MBL_WKS_BOOTLOADER_FS_DEFAULT_SIZE_MiB ?= "10"
+MBL_WKS_BOOTLOADER_FS_SKIP ?= "1"
+MBL_WKS_BOOTLOADER_FS_FSTYPE ?= "vfat"
+MBL_WKS_BOOTLOADER_FS_MOUNT_OPTS ?= "ro,defaults"
+MBL_PARTITION_NAMES += "WKS_BOOTLOADER_FS"
 
 # ------------------------------------------------------------------------------
-# Boot partition
+# BOOT (file system; single bank)
```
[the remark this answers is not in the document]
jonathanhaigh-arm: > Should we mention that this is the partition that contains the kernel, device-tree and initramfs?

Will do.

> Why the BOOT partition is not banked?

Just because I haven't done it yet... I'm thinking that I should probably have a separate PR for:
* Adding the WKS_BOOTLOADER_FS partition.
* Banking the BOOT partition.

Actually using the WKS_BOOTLOADER_FS partition for rpi3 won't work without fiddling with our U-Boot scripts and I don't want to make this PR too big.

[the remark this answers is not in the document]
jonathanhaigh-arm: > Should we mention that this is the partition that contains the kernel, device-tree and initramfs?

https://github.com/ARMmbed/meta-mbl/pull/617/commits/34e632c489ddce4f816e5f19992c7b47b7c6432a

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I've added docstrings and in-line comments in https://github.com/ARMmbed/meta-mbl/pull/617/commits/34e632c489ddce4f816e5f19992c7b47b7c6432a.

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I've got rid of the class in https://github.com/ARMmbed/meta-mbl/pull/617/commits/34e632c489ddce4f816e5f19992c7b47b7c6432a. The methods have been refactored and are now free functions. The data for each partition is now just a `dict`. I'm still not convinced I've got it right... Let me know what you think.

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I've updated the error message when the size in KiB isn't 1024 times the size in MiB: https://github.com/ARMmbed/meta-mbl/pull/617/commits/a49612f990f18d056be2eff1b8a0d30180f11dce

meta-mbl-distro/classes/mbl-partitions.bbclass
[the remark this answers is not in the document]
jonathanhaigh-arm: > I'm thinking that I should probably have a separate PR for...Adding the WKS_BOOTLOADER_FS partition.

https://github.com/ARMmbed/meta-mbl/pull/617/commits/25e076d83d21610aa4390757cc5182af04ff9521, https://github.com/ARMmbed/meta-mbl/pull/617/commits/478648f56ea6c6fe314c11e8639efd3f56955eeb

meta-mbl-distro/recipes-core/packagegroups/packagegroup-mbl-development.bb
jonathanhaigh-arm: Removed mmc-utils in https://github.com/ARMmbed/meta-mbl/pull/617/commits/b050517701a513a3797a1b562a3e16191671b23f.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -9,21 +9,45 @@ SRC_URI_append = " file://fstab "
...
+            part["mount_opts"],
+            part["fs_freq"],
+            part["fs_passno"]
+        )
+        d.appendVar("FSTAB_LINES", fstab_line)
 }
 
 do_install_append() {
     # Ensure that mountpoints specified in fstab exist on the root filesystem
     for mountpoint in ${MOUNT_POINTS}; do
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/617/commits/b073b8cb97133ea6a94d1c0c2082aa97444b4b1a

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -9,21 +9,45 @@ SRC_URI_append = " file://fstab "
 # override hostname in build time. hostname will be set by hostname.sh init script
 hostname_pn-base-files = ""
 
-# Create a list of mount points that need to be created
-python() {
-    for part in d.getVar("MBL_PARTITIONS").split():
-        mountpoint = d.getVar("MBL_{}_MOUNT_POINT".format(part))
-        if mountpoint and mountpoint != "/":
-            d.appendVar("MOUNT_POINTS", " {}".format(mountpoint))
+python __anonymous() {
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/617/commits/b073b8cb97133ea6a94d1c0c2082aa97444b4b1a

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -10,152 +10,456 @@ inherit mbl-artifact-names
 # recipes-core/base-files/files/fstab.
```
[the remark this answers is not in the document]
jonathanhaigh-arm: Yes. The user shouldn't have to touch either the wks file or the fstab recipe.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -10,152 +10,456 @@ inherit mbl-artifact-names
...
+# ------------------------------------------------------------------------------
 # We have two different flavours of "partition" in MBL:
 # * Normal file system partitions, recorded in the partition table.
 # * Raw non-file system "partitions", not recorded in the partition table.
+#
+# For each of MBL's "partitions" there are the following variables:
+# * MBL_<partition>_ALIGN_KiB   - minimum alignment in KiB
+# * MBL_<partition>_SIZE_KiB    - size in KiB
+# * MBL_<partition>_SIZE_MiB    - size in MiB
+# * MBL_<partition>_OFFSET1_KiB - offset of first bank in KiB
```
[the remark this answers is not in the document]
jonathanhaigh-arm: a85de9e46b03b3d7f700c1287e4854a3bb7dccdd

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -10,152 +10,456 @@ inherit mbl-artifact-names
...
+# * MBL_<partition>_LABEL       - label of partition
+# * MBL_<partition>_FSTYPE      - type of file system
+# * MBL_<partition>_MOUNT_OPTS  - mount options
+# * MBL_<partition>_FS_FREQ     - fifth fstab(5) field
+# * MBL_<partition>_FS_PASSNO   - sixth fstab(5) field
+#
+# Notes:
+# * Note that partitions that are are banked (for an A/B firmware update
+#   scheme) have a single set of variables - only the offset variable is
+#   duplicated.
```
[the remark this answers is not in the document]
jonathanhaigh-arm: You're correct. I've reworded: https://github.com/ARMmbed/meta-mbl/pull/617/commits/a668c659278a47c1726b566c61365446f00374cb

meta-mbl-distro/classes/mbl-partitions.bbclass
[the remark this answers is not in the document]
diego-sueiro: Right. So I think we should change this comment.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -10,152 +10,456 @@ inherit mbl-artifact-names
...
-MBL_HOME_PARTITION_SIZE_MiB ?= "450"
-MBL_HOME_PARTITION_ALIGN_KiB ?= "${MBL_FLASH_ERASE_BLOCK_SIZE_KiB}"
+MBL_HOME_LABEL ?= "home"
+MBL_HOME_DEFAULT_SIZE_MiB ?= "450"
+MBL_PARTITION_NAMES += "HOME"
 
 
 # ------------------------------------------------------------------------------
-# Raw non-file system "partitions" not recorded in the partition table
+# Variables for default values
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/617/commits/e102d47c2b096722c232fddadc3bab1739b6c868

jonathanhaigh-arm:
@diego-sueiro, I think I've addressed all of your comments now.

meta-mbl-distro/classes/mbl-partitions.bbclass
rwalton-arm: I think it's an improvement. I probably would've preferred to see a the partition data as a class with public attributes rather than a dict. Don't think it's that important though.

meta-mbl-distro/classes/mbl-partitions.bbclass
rwalton-arm: Thanks, I've been trying to think of a way to drop this `bool(int(val))` but I can't really think of a good reason. `int(" ")` will throw an exception though so you may want to handle that.

jonathanhaigh-arm: I'm not too bothered about it throwing an exception. I just want to avoid it silently doing the wrong thing.

PR #621, opened 2019-07-25
meta-mbl-distro/mbl-initramfs-init: Make sure "/" is always owned by root

diego-sueiro:
When investigating an issue in the test farm for the avahi discovery
test, it was identified that the following services fail to start:
```
    systemd-hostnamed.service
    systemd-logind.service
    systemd-resolved.service
    systemd-timedated.service
    systemd-timesyncd.service
```

The main reason is that these services use the systemd PrivateTmp
feature which expects the /var/tmp to be in place. The systemd-resolved
is the service responsible for dealing with the mDNS responder.

In MBL the /var/tmp is a symlink to /var/volatile/tmp which is created
by systemd-tmpfiles-setup.service by reading the
/etc/tmpfiles.d/00-create-volatile.conf file.

In the failure condition, this is the stripped output for
"journalctl -r --no-pager -l -u systemd-tmpfiles-setup":
`    systemd-tmpfiles[223]: Detected unsafe path transition / -> /var during canonicalization of /var/volatile.`

The root cause of this problem is that "/" is not owned by root, as
we can confirm with the command "stat /"
```
    File: /
    Size: 1024 Blocks: 2 IO Block: 1024 directory
    Device: b303h/45827d Inode: 2 Links: 23
    Access: (0755/drwxr-xr-x) Uid: ( 1000/ UNKNOWN) Gid: ( 1000/ UNKNOWN)
```

To prevent this from happening we set the ownership of "/" to root right
after mounting it.


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/66/

PR #626, opened 2019-07-29
Fix u-boot menuconfig inside mbl-tools interactive mode

diego-sueiro:
In mbl-tools interactive mode, the libncurses5-dev is not installed in the container build image and u-boot menuconfig fails to run.

Apply the same fix as we have for the imx Linux kernel recipes (PR #401).


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/67/

PR #632, opened 2019-07-31
signing: Add signing REST API and backends.

rwalton-arm:
The code signing tools will use a REST API to communicate with the
signing engine. This will allow users to plug in any secrets storage
engine they like. To accomplish this a user must implement a
'SigningBackend' interface and plug that into the API endpoint handler
code.

User defined storage backends are only partly supported in the
BETA version of the API, to fully support this we need to add a mechanism
for selecting the current signing backend.

* backends.py contains the SigningBackend interface and a reference
implementation around a Hashicorp Vault Client.

* endpoints.py is a flask app which handles the HTTP requests to the
signing API.

* keystore.yaml is a swagger 2.0 definition of the API endpoints.

* keystore.py is the interface to the signing tools, which sends HTTP
requests to endpoints.py over a local server.

Jira: IOTMBL-2315


loverdeg-ep:
@ARMmbed/team-embeddedplanet

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
```diff
@@ -0,0 +1,257 @@
...
+        Generate a root CA.
+
+        :params common_name str: Common name for the CA.
+        :params params dict: Optional config params to pass to Vault.
+        """
+        self._client.sys.enable_secrets_engine(
+            backend_type="pki", path=common_name
+        )
+        self._client.sys.tune_mount_configuration(
+            max_lease_ttl=params.get("ttl", "8750h"), path=common_name
```
jjohnson-arm: Does this mean the root CA will only last a year?  Does this need to be configurable?
Or is this just the mount to the secrets backend, if so why do we need a year?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
```diff
@@ -0,0 +1,257 @@
...
+    @abstractmethod
+    def list_certificates(self, *args, **kwargs):
+        """List certificates stored in the backend."""
+
+
+class VaultBackend(SigningBackend):
+    """
+    Implementation of the SigningBackend interface.
+
+    The backend used in this case is Hashicorp Vault.
```
jjohnson-arm: I think it might be worth explaining the mount_point system (as I presume this is Vault specific).
Though - this is probably what you have written up in your document explaining the flow, so maybe its not needed here.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
```diff
@@ -0,0 +1,257 @@
...
+    """
+
+    def __init__(self, server_url=None):
+        """
+        Initialise the Vault client.
+
+        :params server_url str: Vault server url.
+        """
+        self._client = hvac.Client(url=server_url)
```
jjohnson-arm: Is it worth double-checking the client version?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/keystore.py
```diff
@@ -3,100 +3,178 @@
...
+        if hasattr(self, "status") and self.status != 200:
+            msg = "Returned status: {}\n\nError message from server: {}\n"
+            raise KeyStoreError(msg.format(self.status, self.detail))
+
+
+def run_server_proc():
+    """Run the REST api server as a subprocess."""
+    proc = multiprocessing.Process(target=endpoints.main)
+    proc.start()
+    time.sleep(1)
```
jjohnson-arm: Is this a pause to wait for the process to begin?  Or a work-around for an issue?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def main():
+    """Entry point."""
+    # This is needed because global variables don't work in python without it.
+    global engine
+    engine = signing_backend()
+    app = connexion.FlaskApp("keystore-api-server", specification_dir=SPEC_DIR)
+    app.add_api("keystore.yaml", arguments={"title": "KeyStore"})
+    app.run(port=5000)
```
jjohnson-arm: Future configurable?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+"""
+
+import connexion
+import pathlib
+from signing.pki.backends import signing_backend
+
+
+SPEC_DIR = pathlib.Path(__file__).parent / "swagger"
+
+# Set the signing engine to None as it needs to be instantiated
```
diegorusso: We can use the flask `g` object for make this available to the whole flask app. More info: https://flask.palletsprojects.com/en/1.1.x/appcontext/

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/keystore.py
rwalton-arm: It's a pause to wait for the process to begin and the server to start up. Not the nicest solution admittedly, I will add a task to change it in the next version.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/endpoints.py
rwalton-arm: In the future it won't be running the default flask dev server, will use nginx or similar. When I rework this code I'll make it configurable.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/endpoints.py
rwalton-arm: I'll take a look..

rwalton-arm: After discussion with Diego, this is OK for the BETA version, as we're only handling a single HTTP request at a time. I have created https://jira.arm.com/browse/IOTMBL-2418 to cover future API work. I'll also add a comment to the code explaining why we aren't using the flask `g` object.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/endpoints.py
rwalton-arm: Added https://jira.arm.com/browse/IOTMBL-2418 to track future improvements to the API.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
rwalton-arm: I'm attempting to abstract away the 'mount' system in my API. The "front end" of the API doesn't mention mount points anywhere (or shouldn't!). I'm only using that terminology in the "backend" example. I will explain how Vault works in the docs though.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
rwalton-arm: It's the max ttl the secrets backend is capable of. It is configurable, a year is the default value.

https://docs.python.org/3/library/stdtypes.html#dict.get

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
```diff
@@ -0,0 +1,257 @@
...
+
+"""
+This module defines a the "backend" interface for secrets storage engines.
+
+The interface allows the storage engine to be called by the signing REST API.
+
+If you want to configure your own secrets backend, you must implement the
+SigningBackend interface.
+
+However this is not fully supported in this version of the API.
```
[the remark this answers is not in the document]
rwalton-arm: Probably not. I'll remove the weasel words from the docstring.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/backends.py
rwalton-arm: Not sure it is. The client version isn't the same as the Vault server version so I'm not sure what checking the version would gain us.

PR #633, opened 2019-07-31
signing: Add x509 and ASN1 parsing utils.

rwalton-arm:
The signing tools patch x509v3 extensions into the certificates included
in the FIP image. These utilities are used to format the ASN1 data and
create the extensions. The encoded data is then passed to the cert generation
engine to be included in the certificates.

This patch also includes a module which defines data structures
representing the certificates used in the ATF Chain of Trust.

Part of Jira: IOTMBL-2321


openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
```diff
@@ -0,0 +1,315 @@
...
+        :params data bytes: File data in bytes.
+        """
+        if data is None:
+            digest = b"\x00" * 32
+        else:
+            hasher = hashes.Hash(hashes.SHA256(), backend=default_backend())
+            hasher.update(data)
+            digest = hasher.finalize()
+        enc_val = asn1parse.encode_asn1_digest_info(
+            "2.16.840.1.101.3.4.2.1", digest
```
jjohnson-arm: Where does "2.16.840.1.101.3.4.2.1" come from?

jjohnson-arm: Oh I found it in the OID repository - http://oid-info.com/get/2.16.840.1.101.3.4.2.1

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
```diff
@@ -0,0 +1,315 @@
...
+        for i in self._extensions.values():
+            yield i
+
+
+class ExtensionValueGetter:
+    """
+    This thing deduces the extension value type.
+
+    It does this based on the extension object's type.
+    This is an abominable hack and should be removed.
```
jjohnson-arm: Do we have an idea of what we should be doing instead? Do we have a ticket/task for this?
It would be better to explain this is a work-around until we have implemented X.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
```diff
@@ -0,0 +1,315 @@
...
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+
+
+def add_extensions(cert_pem, encoding, extensions):
+    """
+    Add extension objects to a certificate object.
+
+    :params cert_pem str: PEM encoded certificate.
+    :params encoding str: Output cert encoding DER or PEM.
```
jjohnson-arm: or RAW

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/tbbr_defs.py
```diff
@@ -0,0 +1,248 @@
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
```
diegorusso: Can we specify that this comes from https://github.com/ARM-software/arm-trusted-firmware/blob/9d068f66b15e644df4961b74b965323c20f21f14/include/tools_share/tbbr_oid.h ?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
```diff
@@ -0,0 +1,315 @@
...
+        super().__init__(*args, **kwargs)
+        self.pub_key_name = pub_key_name
+
+    def encode_value(self, pub_key):
+        """
+        Encode a subject public key and store in the extension.
+
+        :params pub_key bytes: The raw subject public key bytes.
+        """
+        enc_val = asn1parse.encode_asn1_subject_key_info(
```
diegorusso: do we care if pub_key is None?

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/tbbr_defs.py
jjohnson-arm: Probably not that exact version though possibly...

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/asn1parse.py
```diff
@@ -0,0 +1,70 @@
...
+    :params oid string: Dotted string ObjectIdentifier.
+    :params digest bytes: A digest to include in the ASN1 structure.
+    """
+    digest_info = DigestInfo()
+    digest_info["digestAlgorithm"]["algorithm"] = oid
+    digest_info["digest"] = digest
+    return der_encoder(digest_info)
+
+
+def encode_asn1_subject_key_info(oid, pub_key):
```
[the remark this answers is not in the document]
rwalton-arm: We don't. Forgot to remove it after I refactored this. Not needed because the `SubjectPublicKeyInfo` produced from py-cryptography when extracting the public key from the certificate already contains the OID embedded in the data.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
rwalton-arm: I'm planning on improving this code when I come to work on the release version of the signing tool. I'll add it as a point to one of the tasks. It's not really a work around, it's just a bit rubbish, but I don't have time to come up with anything better than this right now.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
rwalton-arm: `pub_key` is never `None` here, that's handled in the `ExtensionValueGetter`. If `pub_key` is `None` it returns \x00*32. This isn't going to change for the BETA.

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
rwalton-arm: Changed the docstring in 712d58f

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/tbbr_defs.py
rwalton-arm: 712d58f

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/asn1parse.py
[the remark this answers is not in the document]
rwalton-arm: 712d58f

openembedded-core-mbl/scripts/lib/mbl-signing/signing/pki/x509utils.py
rwalton-arm: 18546cc

PR #639, opened 2019-08-05
meta-mbl-distro: move mbl-uboot-sign.bbclass to openembedded-core-mbl

diego-sueiro:
Having mbl-uboot-sign.bbclass in meta-mbl-distro is not correct since
it adds an unnecessary dependency on meta-mbl-distro for bsp specific
layers. Hence this class needs to be moved to openembedded-core-mbl
layer.


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/73/

PR #641, opened 2019-08-06
Update partition table

jonathanhaigh-arm:
* Add a second boot partition bank to support A/B updates of kernel
  images.

* Remove the bootflags partition - we plan to replace this with
  bank/update state in raw flash storage so that it is accessible from
  early bootloaders. Until we add support for accessing bank/update
  state from raw flash storage though, we'll still need the boot flag
  files to determine which root file system to use, so move them to
  /var/log for now.

* Add a "bootloaderfs" partition on Raspberry Pi 3 to hold VC4 firmware,
  our BL1 shim and BL2. We need the bootloaderfs partition to be
  separate from the boot partitions so that updates of the boot
  partitions won't affect the contents of the bootloaderfs partition.

  Currently, on Raspberry Pi 3, U-Boot just looks in the first FAT file
  system partition for the FIT image, but now it should be looking in
  the first boot partition rather than the bootloaderfs partition. This
  is unfortunate, but until we fix it, just put everything (the VC4
  firmware, BL1 shim, BL2 and the FIT image) into both partitions.
  We'll have to be careful about how we do FIT updates...

* Don't populate the second root partition with files when creating a
  Wic image - we don't need to and it makes flashing the image take
  longer (at least when using bmaptool).

For:
* IOTMBL-2395: Update partition layout - remove bootflags partition
* IOTMBL-2393: Update partition layout - Add a second boot partition bank
* IOTMBL-2394: Update partition layout - Add an optional bootloader file
  system for e.g. RPi3

**Jenkins Links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/132/ (succeeded).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/134/ (Failed in xtest's do_compile; after rootfs update fix).
* http://jenkins.mbed-linux.arm.com/job/jh-test/137/ (succeeded).

**Lava Links**
* http://e120856-lin/results.php?image_version=custom&image_number=132&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. Rootfs updates failed.
* http://e120856-lin/results.php?image_version=custom&image_number=137&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job (After addressing review comments).
  * A Wi-Fi test failure on Warp7 that succeeded on a second try.
  * An avahi-discovery failure during rootfs update on Warp7 that succeeded on a second try.
  * Neither of these failures seem related to partitioning changes. 

**Notes**
* ~With this PR, the partition number of the first root file system changes on Raspberry Pi 3 from 3 to 5. Our Lava infrastructure will need updating for this.~ The root file system partition number has only changed on Raspberry Pi 3, and, conveniently, on that platform, Lava doesn't need to know the root file system partition number so no changes are required.


jonathanhaigh-arm:
The rootfs update Lava jobs failed - I think I need to fix two things:
* Ensure that the bootflags directory exists.
* Ensure that the U-Boot scripts look in the right place for the flags.

jonathanhaigh-arm:
@diego-sueiro, @rwalton-arm, I've made changes to fix rootfs update (after I broke it). Can you take a look please?

meta-mbl-apps/recipes-connectivity/mbl-cloud-client/mbl-cloud-client.bb
```diff
@@ -139,6 +140,8 @@ do_install() {
 
     install -d "${D}${systemd_unitdir}/system/"
     install -m 0644 "${WORKDIR}/mbl-cloud-client.service" "${D}${systemd_unitdir}/system/"
+
+    install -d "${D}${MBL_BOOTFLAGS_DIR}"
```
diego-sueiro: Why do we need to add an empty directory in the package?

meta-mbl-distro/recipes-core/mbl-initramfs-init/files/initramfs-init-script.sh
```diff
@@ -64,20 +70,24 @@ echo "Booting from init script in initramfs"
...
+ROOTFS_LABEL="${ROOTFS_LABEL_BASE}2"
+if [ ! -f "${BOOTFLAGS_DIR}/${ROOTFS_LABEL}" ]
 then
-    ROOTFS_LABEL="rootfs1"
+    ROOTFS_LABEL="${ROOTFS_LABEL_BASE}1"
 fi
 
+# We'll mount the log partition again after we've done the switch_root so don't
+# leave it hanging around here.
+umount "$LOG_PARTITION"
```
diego-sueiro: I'm not sure if I agree with this operation. Systemd is smart enough to mount what needs to be mounted.
Unless we identify a potential security risk by leaving it mounted.

And I know that it is not part of this PR, but maybe we could think about logging initramfs errors into the log partition.

meta-mbl-apps/recipes-connectivity/mbl-cloud-client/mbl-cloud-client.bb
jonathanhaigh-arm: So that it exists when the update scripts attempt to write flag files there.

We _could_ have the update scripts create the directory if it doesn't exist, but I don't really want to do that - I want them to be agnostic about how the directory is created (whether it's a simple `mkdir` or whether it requires a `mount`).

...Or was that a rhetorical question because you think I should add a code comment? I'll add a code comment.

diego-sueiro: Well, I don't think that makes sense the recipe packing an empty directory. I think that the source code responsibility to handle the dependencies, in this case creating a directory, if they don't exist.

meta-mbl-distro/recipes-core/mbl-initramfs-init/files/initramfs-init-script.sh
jonathanhaigh-arm: I will make the comment more clear... I didn't add the `umount` just because we'll try to mount it again later, I added the `umount` because we'll try to mount it again later _at a different directory_.

We mount it here at `$MBL_LOG_MOUNT_POINT`, then we do the `switch_root`, then systemd will try to mount it again at `$MBL_LOG_MOUNT_POINT`, but after the `switch_root`, the path `$MBL_LOG_MOUNT_POINT` does not refer to the same directory as it did before.
```
root_before_switch_root
 |-var
    |-log ($MBL_LOG_MOUNT_POINT before switch_root)
 |-mnt
    |-rootfs (root_after_switch_root)
       |-var
          |-log ($MBL_LOG_MOUNT_POINT after switch_root)
```

It would be nice to just mount the log partition at `/mnt/rootfs/var/log` in the initramfs script and leave it mounted, but we need to mount the log partition to check the boot flags before we can mount the proper rootfs at `/mnt/rootfs`.

jonathanhaigh-arm: I suppose we could potentially do
```
mount --move "${MBL_LOG_MOUNT_POINT}" "/mnt/root${MBL_LOG_MOUNT_POINT}"
```

diego-sueiro: Right. In this case, you can use `mount --move /var/log /mnt/rootfs/var/log` like we do for `/dev/`, `/proc` and `/sys`.

diego-sueiro: I added my previous comment without refreshing the page. I'd prefer to use the `mount --move` instead of unmounting and then mounting again with systemd.

meta-mbl-apps/recipes-connectivity/mbl-cloud-client/mbl-cloud-client.bb
jonathanhaigh-arm: > Well, I don't think that makes sense the recipe packing an empty directory.

Why?

For the reasons I stated above, I don't want the source code to create the directory.

jonathanhaigh-arm: Comment added in https://github.com/ARMmbed/meta-mbl/pull/641/commits/474b0cdef7d933b77c7e70226c3ab77c52c4ff39.

meta-mbl-distro/recipes-core/mbl-initramfs-init/files/initramfs-init-script.sh
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/641/commits/474b0cdef7d933b77c7e70226c3ab77c52c4ff39

meta-mbl-apps/recipes-connectivity/mbl-cloud-client/mbl-cloud-client.bb
diego-sueiro: Right, having the knowledge of what kind of operation needs to be performed to have this path in place is a good reason for not embedding it into the code. But the same _problem_ applies at the recipe level. Correct?

Is this `MBL_BOOTFLAGS_DIR` supposed to be used by other recipes/packages/apps? If so, this path needs to be provided by a _common_ recipe like base-files.

jonathanhaigh-arm: > But the same problem applies at the recipe level. Correct?

Sort-of, but less so. How the directory is created is platform and distro specific, and I think it's more reasonable to have that sort of thing in recipes than in source code.

> Is this MBL_BOOTFLAGS_DIR supposed to be used by other recipes/packages/apps?

Mmmm... Good point. I'll create it in base-files.

jonathanhaigh-arm: Moved to base-files in https://github.com/ARMmbed/meta-mbl/pull/641/commits/9275a7d3a84479a25a49da5bd21058908df6fe6c and https://github.com/ARMmbed/meta-mbl/pull/641/commits/67fa872a925d91bc91ca2c56e817c9d2f3023052.

PR #645, opened 2019-08-06
Fix alignments of logical partitions

jonathanhaigh-arm:
Previously, our code to calculate the offsets of partitions didn't take
into account the extended boot records that precede logical partitions.

Add code to make sure that we leave an alignment sized chunk of space
before each logical partition to hold an EBR. The EBRs are only 512B,
but leaving an alignment sized chunk should mean that the EBRs don't
share flash erase blocks with the actual partition data, reducing the
risk of corrupting the partition table if a power failure occurs during
a flash erase or write operation.

The failure to take into account the EBRs meant that some partitions
ended up with double the intended offsets, making our wic images much
bigger than they needed to be. Now that the images are smaller, we don't
need the imx6ul specific partition sizes to cut down on image size, so
just use the defaults on imx6ul.

Also, fix a "bb.warn" line with a missing call to "str.format".

For IOTMBL-2419: mbl-partitions.bbclass: partition alignment
calculations don't take extended boot records into account

**Jenkins links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/133/ (succeeded).

**Lava links**
* http://e120856-lin/results.php?image_version=custom&image_number=133&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. There were two failures:
    * https://lava.mbedcloudtesting.com/scheduler/job/34480: timeout during rootfs update test on imx8mm; investigating. There are 38 jobs in the queue for our imx8mms in the test farm so I haven't re-submitted this one. A lot of the time was spent waiting for a login prompt before timing out. Three times it looks like the timeout was due to RNG init taking too long; another time it looks like something got stuck in the initramfs script...
    * https://lava.mbedcloudtesting.com/scheduler/job/34511: Wi-Fi test failure; succeeded after re-submission (https://lava.mbedcloudtesting.com/scheduler/job/34783).


meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -346,9 +374,13 @@ python __anonymous() {
             bb.fatal("{} is not set".format(offset_var))
         elif have_prev:
             offset = _calculate_next_offset(align, prev_offset, prev_size)
+            if prev_fs_part_number > 2:
```
diego-sueiro: Can you add a comment about what is expected to be the partitions > 2?

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -460,6 +494,11 @@ python __anonymous() {
         part_info["mount_opts"] = _process_var_with_default(d, part_name, "MOUNT_OPTS")
         part_info["fs_freq"] = _process_var_with_default(d, part_name, "FS_FREQ")
         part_info["fs_passno"] = _process_var_with_default(d, part_name, "FS_PASSNO")
+        part_info["fs_part_numbers"] = []
```
diego-sueiro: Can you add a comment about the reason why we are doing this computation?

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -346,9 +374,13 @@ python __anonymous() {
             bb.fatal("{} is not set".format(offset_var))
         elif have_prev:
             offset = _calculate_next_offset(align, prev_offset, prev_size)
+            if prev_fs_part_number > 2:
+                # We need to leave space for an alignment sized chunk for a
+                # 512B extended boot record
+                offset = _calculate_next_offset(align, offset, 1)
```
diego-sueiro: I don't understand why the `prev_size` is 1.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -299,7 +299,26 @@ python __anonymous() {
...
         return size_KiB, size_MiB
 
-    def _process_offset_vars(d, part_name, is_banked, size, align, prev_offset, prev_size):
+    def _next_fs_part_number(fs_part_number):
+        """
+        Given a partition number, return the next non-extended partition
+        number.
+        """
+        if fs_part_number == 3:
+            # Skip a part number to account for the extended partition
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/645/commits/c3be70492abbee4850ceaaef4de7850539d7cc6d

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/645/commits/c3be70492abbee4850ceaaef4de7850539d7cc6d

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/645/commits/c3be70492abbee4850ceaaef4de7850539d7cc6d

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: Mmmm. I'd rather not.

The reasons for the computation are:
1. We'll need it later when telling U-Boot where to find the FIT image to use.
2. It's useful for debug.

There aren't comments for every other field of `part_info`, and I don't think "fs_part_numbers" is special enough to warrant a comment. If you don't think my reasons for having the field are good enough then I should get rid of the field rather than add a comment... Although I'd rather keep it.

PR #657, opened 2019-08-08
meta-mbl-distro: Save partition size, alignment and offsets

rwalton-arm:
Save the raw partition offsets and sizes in the factory
config partition. The firmware update code will need this data to
perform an update of a component.

Add a function to base-files.bbappend to save the partition information.
Add sourceparams to mbl.wks.in to tell it to populate the factory config partition.

Jira: IOTMBL-2397

Jenkins: http://jenkins.mbed-linux.arm.com/job/rw-test/135/

LAVA: http://e120856-lin.cambridge.arm.com/results.php?image_version=custom&image_number=135&custom_version=rw-test&lava_query=&lava_owner=&failures=false&submitter=robwal02&sort=by_job


meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -42,6 +42,27 @@ python __anonymous() {
         d.appendVar("FSTAB_LINES", fstab_line)
 }
 
+# Write the partition sizes and offsets to files which will be placed in /config/factory
+python __anonymous() {
```
diego-sueiro: Can we create a new task and execute it before do _install?
Creating the WORKDIR at parse time is a bit hacky to me.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -57,4 +78,14 @@ do_install_append() {
...
     cat ${D}${sysconfdir}/fstab
+
+    # Install the partition data files in the factory config partition
+    # This information is needed for component updates.
+
+    install -d "${D}${MBL_FACTORY_CONFIG_MOUNT_POINT}/part-infos"
+
+    for file in "${MBL_PARTITION_DATA_FILES}"
+    do
+        install -m 0644 $file "${D}${MBL_FACTORY_CONFIG_MOUNT_POINT}/part-infos"
```
diego-sueiro: I suppose the partition offset information file is ro, correct? If so it should be installed with `0444`.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -42,6 +42,27 @@ python __anonymous() {
         d.appendVar("FSTAB_LINES", fstab_line)
 }
 
+# Write the partition sizes and offsets to files which will be placed in /config/factory
+python __anonymous() {
+    import pathlib
+
+    path_prefix = d.getVar("WORKDIR")
+    # the WORKDIR doesn't exist at recipe parse-time, so create it.
```
jonathanhaigh-arm: Why are doing this at parse time rather than in a task?

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -42,6 +42,27 @@ python __anonymous() {
...
+    data_required = ["align_KiB", "offsets_KiB", "size_KiB"]
+
+    for part_info in d.getVar("MBL_PARTITION_INFOS"):
+        if part_info["skip"]:
+            continue
+
+        for data_item in data_required:
+            fname = "{}_{}".format(part_info["name"], data_item)
+            fpath = pathlib.Path(path_prefix, fname)
+            fpath.write_text(str(part_info[data_item]))
```
jonathanhaigh-arm: What happens here when `data_item` is  `"offsets_KiB"`? `part_info["offsets_KiB"]` is a `list`...

rwalton-arm: It saves a string representation of the list, which ends up being "[item, item2]" etc. I guess I could save each element on it's own line.

jonathanhaigh-arm: Yeah, I think being able to parse the data very easily should be a high priority here.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
jonathanhaigh-arm: I think using `do_compile` would be fine as well. But if `do_compile` has already been defined as a shell function then it would be easier to create a new task.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -42,6 +42,38 @@ python __anonymous() {
         d.appendVar("FSTAB_LINES", fstab_line)
 }
 
+# Write the partition sizes and offsets to files which will be placed in /config/factory
+python do_save_partition_info() {
+    import pathlib
+
+    path_prefix = d.getVar("WORKDIR")
```
diego-sueiro: Can you `s/path_prefix/workdir`? I think it is more clear reading `workdir` and understand what is going on. And looking at the oe-core classes they follow this naming convention.

rwalton-arm:
> Looking good; just one minor comment below.
> 
> Did you work out what was going on with the weird vardeps behaviour you saw?

Not yet, I pushed my changes so I could try running a build in Jenkins to see if there was something wrong with my enviironment - unfortunately I'm still seeing errors like the following
```
ERROR: When reparsing /mnt/workspace/rw-test-2/machine-raspberrypi3-mbl/mbl-manifest/build-mbl/conf/../../layers/openembedded-core/meta/recipes-core/base-files/base-files_3.0.14.bb.do_save_partition_info, the basehash value changed from 0a4aaf5afece69f7a42917d2a29765e841fca76425c851a27a830f5af8c1bec1 to 3c84de7cbfe0d09b8b7c3e9d1940da6443502a1d57263d20e9742482f7a02618. The metadata is not deterministic and this needs to be fixed.
```

It works when I use an anonymous function but not in a task for some reason. I'm going to try and fix it this morning. I've added the DO NOT MERGE label to this PR.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -227,23 +246,26 @@ MBL_LOG_LABEL ?= "log"
...
 # ------------------------------------------------------------------------------
 # Home partition (file system; single bank)
 # ------------------------------------------------------------------------------
 MBL_HOME_MOUNT_POINT = "${MBL_HOME_DIR}"
 MBL_HOME_LABEL ?= "home"
 MBL_HOME_DEFAULT_SIZE_MiB ?= "450"
 MBL_PARTITION_NAMES += "HOME"
-
+MBL_PARTITION_VARS += "\
+MBL_HOME_MOUNT_POINT MBL_HOME_LABEL MBL_HOME_DEFAULT_SIZE_MiB"
```
jonathanhaigh-arm: I think `MBL_PARTITION_VARS` is missing a bunch of variables that are only set in the Python section below.

I suggest just creating `MBL_PARTITION_VARS` in the Python section.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
```diff
@@ -42,6 +44,23 @@ python __anonymous() {
...
+    workdir = pathlib.Path(d.getVar("WORKDIR", True))
+    for var in d.getVar("MBL_PARTITION_VARS", True).split():
+        var_value = d.getVar(var, True)
+        file_path = pathlib.Path(workdir, var)
+        file_path.write_text(var_value)
+        d.appendVar("MBL_PARTITION_FILES_STAGING_DIR", str(file_path))
+}
+
+addtask write_partition_vars before do_install
+do_write_partition_vars[vardeps] += "MBL_PARTITION_VARS"
```
jonathanhaigh-arm: In addition to `MBL_PARTITION_VARS`, you should add `${MBL_PARTITION_VARS}` - each variable needs to be added to vardeps. You've also used `MBL_PARTITION_FILES_STAGING_DIR`... TBH, I think it would be easier to write this function in shell and let BitBake find most of the vardeps.

rwalton-arm: I had an issue writing this function in shell, when I tried to use python inline variable expansion it complained about a bad substitution, I had some code similar to:

```
for var in ${MBL_PARTITION_VARS}; do
    printf "%s" "${@ d.getVar('${var}', True)}" > ${PARTITION_VARIABLES_DIR}/${var}
done
```
Seemed like it didn't like the `${var}` expansion in the inline python expression - which is why I ended up with this.

meta-mbl-distro/classes/mbl-partitions.bbclass
rwalton-arm: Will do.

meta-mbl-distro/recipes-core/base-files/base-files_%.bbappend
jonathanhaigh-arm: Ah, okay.

meta-mbl-distro/classes/mbl-partitions.bbclass
jonathanhaigh-arm: I see this was done in https://github.com/ARMmbed/meta-mbl/pull/657/commits/ecf4db4b2ffd7350ef86cfa170c8e4dc14ecf8bf.

Thanks :-)

PR #660, opened 2019-08-13
imx8mm: Fix kernel hang when init ramfs

niej:
Update kernel SRCREV to fix kernel hang when init ramfs.
Disabling CONFIG_ARCH_FSL_IMX8MQ in kernel config can avoid
change bus frequency that lead to hang if ATF is updated to
V2.1.

Signed-off-by: Jun Nie <jun.nie@linaro.org>


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/82/

diego-sueiro:
@petedyerarm, conducted a test in the farm where after 50 reboot cycles we couldn't reproduce any boot issues.

PR #664, opened 2019-08-14
Various fixes for u-boot append recipes

diego-sueiro:
This PR includes the following changes:

**meta-raspberrypi-mbl: u-boot: Remove redundant merge_config.sh call:**
The u-boot.inc from oe-core already provides a mechanism to call the
merge_config.sh, so we don't need to call it again.

**u-boot: Fix compile_append for incremental builds:**
During u-boot incremental builds we get the following error:
```
|   cp dts/dt.dtb u-boot.dtb
| cp: 'dts/dt.dtb' and 'u-boot.dtb' are the same file
| /<snip>/Makefile:970: recipe for target 'u-boot.dtb' failed
```
To fix this we need to copy the dts/dt.dtb to ${UBOOT_DTB_BINARY}
(which can have a different name from u-boot.dtb) instead of symlinking
to ${UBOOT_DTB_BINARY}.

**meta-freescale-mbl: u-boot-imx: Workaround to use merge_config.sh:**
When setting UBOOT_CONFIG variable the do_configure defined in
u-boot.inc in oe-core doesn't call the merge_config.sh.

We added this workaround to call the merge_config.sh and prevent
patching the machine defconfig file for imx8mmevk-mbl.


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/83/

meta-freescale-mbl/recipes-bsp/u-boot/u-boot-imx_%.bbappend
```diff
@@ -7,18 +7,27 @@ inherit mbl-uboot-sign
...
 # MBL_UBOOT_VERSION should be updated to match version pointed to by SRCREV
 MBL_UBOOT_VERSION = "2018.03"
 
 SRCREV = "e9cb2c6d8a6227a189702ab2cfc7b1273689ddb2"
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/u-boot-imx:"
 
+
+do_configure_prepend_imx8mmevk-mbl() {
+    # Workaround to use merge_config.sh when using UBOOT_CONFIG variable
```
rwalton-arm: Could you expand this comment to explain a bit more why we're adding this workaround?

diego-sueiro: Added in 0cbcfa4

PR #670, opened 2019-08-16
oecore-mbl: Add a script to create update payloads

rwalton-arm:
We need to handle filenames for bootloader components on different
targets in our update scripts. This utility will pack a tar archive with
the update components and handle renaming the files to generic names
for use by the update scripts on the target.


openembedded-core-mbl/scripts/contrib/create-update-payload.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
```
diego-sueiro: I think this script should be located in `meta-mbl-distro/scripts` since it is MBL distro specific and I don't think we have plans to submit it to oe-core.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+
+"""
+This script will create a bootloader update payload tar archive.
+
+You can specify a compression type for the tar archive. The default is gz and
+the available options are gz, bz2 and xz.
+
+To use the script you can either pass in a space separated list of file names
+or a list of bootloader slot numbers. We then map the file names of the update
+components to the corresponding bitbake variables, which are named according
```
jonathanhaigh-arm: I'm not sure the stuff about BitBake variables should be in this comment - it's an implementation detail. Perhaps explaining what the "bootloader slot numbers" are would be more useful.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+"""
+This script will create a bootloader update payload tar archive.
```
jonathanhaigh-arm: What about rootfs updates and app updates?

I'd be happy for you to say that they're not within the scope of this PR/task, but I don't think we should limit this comment to bootloaders.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+any MACHINE target.
+
+The script assumes the files you want to archive are in DEPLOY_DIR_IMAGE in
+your MBL workspace.
+
+The script supports multi-component or single component payloads.
+
+Example invocation:
+
+create-update-payload --bootloader-slots 1 2 --output-tar-path tmp/payload.tar
```
jonathanhaigh-arm: How does the script know where to find the bootloader files? Do I have to be in a particular directory to run this command? Or do I have to have sourced `setup-environment`?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+import shutil
+import sys
+import tarfile
+import tempfile
+
+from itertools import product
+
+# Set the PYTHONPATH so we can import openembedded scripts/lib modules.
+# This script can only be executed from the
+# meta-mbl-distro/scripts folder in the MBL workspace.
```
jonathanhaigh-arm: I think the note about from where the script can be executed should be in the file's docstring along with the sample invocation.

jonathanhaigh-arm: ...But hang on, isn't `__file__` (used below) the path to the current Python file? Then can't we just work out where things are relvative to this file rather than saying we must execute the script from a particular directory?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+        type=lambda x: pathlib.Path(x).absolute(),
+        default="payload.tar",
+    )
+    parser.add_argument(
+        "-c", "--compression-format", default="gz", choices=["gz", "bz2", "xz"]
+    )
+    args = parser.parse_args()
+
+    if args.filenames is None and args.bootloader_slots is None:
+        raise Exception()
```
jonathanhaigh-arm: Is there some magic going on that means we'll get a reasonable error message here? If not, can we have a reasonable error message please?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+    file names as values.
+
+    :param filenames list: List of file names to look for.
+    :param tinfoil Tinfoil: bitbake tinfoil object.
+    """
+    return {
+        cvar.split("_FILENAME")[0]: fname
+        for cvar, fname in product(tinfoil.config_data, filenames)
+        if cvar.startswith("MBL_WKS")
+        and tinfoil.config_data.getVar(cvar) == fname
```
jonathanhaigh-arm: Should we `strip()` the result of `getVar()`?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+
+    Return a dict with the variable names as keys and
+    file names as values.
+
+    :param filenames list: List of file names to look for.
+    :param tinfoil Tinfoil: bitbake tinfoil object.
+    """
+    return {
+        cvar.split("_FILENAME")[0]: fname
+        for cvar, fname in product(tinfoil.config_data, filenames)
```
jonathanhaigh-arm: There are a lot of `cvar`s here, right? Should we filter for starting with "MBL_WKS" before we create the product?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+    MBL_WKS_BOOTLOADER variable to the given bootloader slot.
+
+    Return a dict with the variable name as the key and the file name as the
+    value.
+
+    :param bootloader_slots list: List of bootloader slots e.g [1, 2, fs]
+    :param tinfoil Tinfoil: bitbake tinfoil object.
+    """
+    output_map = {}
+    for conf_var, bl_slot in product(tinfoil.config_data, bootloader_slots):
```
jonathanhaigh-arm: Again, there will be a lot of `conf_var`s here, right? I think something like the following would be simpler and avoid the massive product iterator.
```Python
for conf_var in tinfoil.config_data:
    m = re.match( r"MBL_WKS_BOOTLOADER_?([A-Z0-9]+)_FILENAME", conf_var)
    if m and m.group(1).lower() in bootloader_slots:
        # ...

File: meta-mbl-distro/scripts/create-update-payload [hunk 10]
```diff
@@ -0,0 +1,191 @@
...
+    :param filename_var_map dict: Mapping of bitbake variable names to files.
+    :param staging dir Path: path to the dir where the files will be copied to.
+    """
+    deploy_dir_img = pathlib.Path(
+        tinfoil.config_data.getVar("DEPLOY_DIR_IMAGE")
+    )
+    for var, fname in filename_var_map.items():
+        src = deploy_dir_img / fname
+        dest = staging_dir / var
+        shutil.copy(str(src), str(dest))
```
jonathanhaigh-arm: Do we need to be careful about copying metadata here? I'm not sure.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,191 @@
...
+
+
+def create_payload_tar_archive(tar_archive_name, dir_path, compression_format):
+    """
+    Create the payload tar archive.
+
+    :param tar_archive_name str: Path to the tar archive.
+    :param dir_path Path: Path to the directory of the files to archive.
+    :param compression_format str: The compression format. Valid options are
+    gz, bz2 and xz.
```
jonathanhaigh-arm: Like I wrote above, we don't want to compress the tar archive, but we probably do want to compress the files inside it... And like above, I'm not sure we should be offering so many compression formats... What does the update script support? For rootfs files it only supports `xz` I think.

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: Ah, that's just me being sloppy. I had a mental note to remove this, raise a specific exception and add an error msg, but I forgot - I'll change this.

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: The comment is wrong, we can execute the script from anywhere, but we can't move the script itself from the folder it's currently in as that will screw up the import paths. The problem is the script needs to import stuff from openembedded core to use the bitbake and oe libs. I'll move the comment to the docstring and correct it.

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: The script looks for the files in `DEPLOY_DIR_IMAGE` so I think we need to have sourced `setup-environment`

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: I think rootfs and app updates need some thought and are probably out of scope for this. I'll edit the docstring so it's not limited to bootloaders.

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: Hmm, yeah we probably should be more careful about copying the flags/permission bits etc. I'll change this to a `copy2`

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: I'll remove the compression for the tar archive and compress the files with `xz`. I'll also remove the other options for now.

rwalton-arm:
Review comments addressed in b5cfd38

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Look up the bitbake variables corresponding to the bootloader slots.
+
+    Parse the bitbake config data using the tinfoil API and match the
+    MBL_WKS_BOOTLOADER variable to the given bootloader slot.
+
+    Return a dict with the variable name as the key and the file name as the
+    value.
+
+    :param bootloader_slots list: List of bootloader slots e.g [1, 2]
```
jonathanhaigh-arm: Ah, I notice you've removed "fs" from that list. I was just thinking that we can't handle "fs" like the others because it doesn't have a `_FILENAME` variable - it uses OE's BOOTFILES mechanism.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,214 @@
...
+
+Example invocation:
+
+create-update-payload --bootloader-slots 1 2 --output-tar-path tmp/payload.tar
+
+This will create the update payload containing the bootloader partition 1 and
+bootloader partition 2 components at tmp/payload.tar.
+
+The script looks for the files in DEPLOY_DIR_IMAGE, you need to have an MBL
+workspace set up and source setup-environment.
```
jonathanhaigh-arm:
```suggestion
workspace set up and you must have sourced setup-environment.
```

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -0,0 +1,214 @@
...
+bitbakepath = scriptpath.add_bitbake_lib_path()
+
+import bb.tinfoil
+import argparse_oe
+
+
+def parse_args():
+    """Parse the command line arguments."""
+    parser = argparse_oe.ArgumentParser()
+    group = parser.add_mutually_exclusive_group(required=True)
```
jonathanhaigh-arm: Much nicer solution :-)

PR #671, opened 2019-08-19
Partition table: Reserve space for bank/update state

jonathanhaigh-arm:
When we implement A/B update of BL3, BL2 will need access to state that
tells it which bank of BL3 to boot. BL2 doesn't understand file systems
so reserve some space for bank/update state in raw flash storage.

For IOTMBL-2396: Update partition layout - Add space for bank/update
state

**Jenkins links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/141/ (succeeded).

**Lava links**
* http://e120856-lin/results.php?image_version=custom&image_number=141&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job (all passed).


meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -163,6 +163,22 @@ MBL_WKS_BOOTLOADER2_IS_BANKED = "1"
 MBL_WKS_BOOTLOADER2_NO_FS = "1"
 MBL_PARTITION_NAMES += "WKS_BOOTLOADER2"
 
+# ------------------------------------------------------------------------------
+# BANK_AND_UPDATE_STATE (non-file system; 1 bank)
```
diego-sueiro: Should we follow the standard and have `single` instead of `1`?

jonathanhaigh-arm: Yes, we should: https://github.com/ARMmbed/meta-mbl/pull/671/commits/b1585c233087454720525b2276eec828a7780681.

PR #672, opened 2019-08-19
meta-mbl-distro: Add root password setting for production images

diego-sueiro:
For production images we want to set the root password stored as plain
text from the file listed in the MBL_ROOT_PASSWD_FILE variable.

The MBL_MIN_PASS_LEN holds the requirement for the minimum password
length.

Other requirements (ENCRYPT_METHOD, FAIL_DELAY, LOGIN_RETRIES and
FAILLOG_ENAB) are directly set in the shadow_%.bbappend.

The shadow-native package had to be patched for checking MIN_PASS_LEN
config when setting the user password in plain text.


diego-sueiro:
Implements: https://jira.arm.com/browse/IOTMBL-2374
**NOTE:** To be completed this feature still needs changes in mbl-tools.

meta-mbl-distro/classes/mbl-extrausers.bbclass
```diff
@@ -0,0 +1,41 @@
+# Copyright (c) 2018 Arm Limited and Contributors. All rights reserved.
```
diego-sueiro:
```suggestion
# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
```

meta-mbl-distro/classes/mbl-extrausers.bbclass
```diff
@@ -0,0 +1,41 @@
...
+
+MBL_ROOT_PASSWD_FILE ?= "${TOPDIR}/mbl_root_passwd_file"
+
+# At parsing time we need to check if the file holding the root password
+# exists and if it meets the requirements set by the MBL_MIN_PASS_LEN
+# variable.
+python __anonymous () {
+    mbl_root_passwd_file = d.getVar('MBL_ROOT_PASSWD_FILE', True)
+    if not os.path.exists(mbl_root_passwd_file):
+        bb.fatal ("MBL_ROOT_PASSD_FILE {} doesn't exist".format(d.getVar('MBL_ROOT_PASSWD_FILE', True)))
```
jonathanhaigh-arm:
```suggestion
        bb.fatal ("MBL_ROOT_PASSWD_FILE \"{}\" doesn't exist".format(mbl_root_passwd_file))
```

meta-mbl-distro/classes/mbl-extrausers.bbclass
```diff
@@ -0,0 +1,41 @@
...
+    elif not os.path.getsize(mbl_root_passwd_file):
+        bb.fatal ("MBL_ROOT_PASSD_FILE {} is empty".format(d.getVar('MBL_ROOT_PASSWD_FILE', True)))
```
jonathanhaigh-arm:
```suggestion
        bb.fatal ("MBL_ROOT_PASS"D_FILE \"{}\" is empty".format(mbl_root_passwd_file))
```

meta-mbl-distro/classes/mbl-extrausers.bbclass
```diff
@@ -0,0 +1,41 @@
...
+
+    try:
+        with open(mbl_root_passwd_file, 'r') as f:
+            mbl_root_passwd = f.readline().strip()
+    except (IOError, OSError) as exc:
+        bb.fatal ("Fail to open {} file".format(mbl_root_passwd_file))
```
jonathanhaigh-arm: TBH, I'm wondering whether all of the error handling here and above is worth it - if you just do the `open` (without the `try` block and the checks above) then if there's a problem the build will fail with an exception, and the exception details should tell the user what went wrong.

Okay, the checks above will cause better error messages for the cases they check, but the error message for when there's an exception in this `try` block is not very useful - it doesn't say why we failed to open the file... And what about if the `readline` fails? Will we get an incorrect "Fail to open {} file" message?

I'd prefer to just get rid of all the checks and let any exceptions propogate, but if you really don't want to do that then I suggest something like:
```suggestion
        bb.fatal ("Failed to read root password from file \"{}\": {}".format(mbl_root_passwd_file, exc))
```
Although I'm not sure that that's the best way to print the info for the original exception. https://stackoverflow.com/questions/4308182/getting-the-exception-value-in-python suggests that `repr()` might be the way to go.

meta-mbl-distro/classes/mbl-extrausers.bbclass
```diff
@@ -0,0 +1,41 @@
...
+    else:
+        if len(mbl_root_passwd) < int(d.getVar('MBL_MIN_PASS_LEN', True)):
+            bb.fatal ("MBL_ROOT_PASSWORD minimum lenght is {}".format(d.getVar('MBL_MIN_PASS_LEN', True)))
```
jonathanhaigh-arm:
```suggestion
            bb.fatal ("MBL_ROOT_PASSWORD minimum length is {}".format(d.getVar('MBL_MIN_PASS_LEN', True)))
```

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-License-Identifier: MIT
+
+MBL_MIN_PASS_LEN ?= "5"
+
+do_install_append () {
+    # We want to use SHA512 as encryption method to generate passwords for the target
+    sed -i 's/^#ENCRYPT_METHOD.*$/ENCRYPT_METHOD SHA512/' ${D}${sysconfdir}/login.defs
+
+    # We want the minimum password length set by MBL_MIN_PASS_LEN for the target
+    sed -i 's/^PASS_MIN_LEN.*$/PASS_MIN_LEN ${@d.getVar('MBL_MIN_PASS_LEN', True)}/' ${D}${sysconfdir}/login.defs
```
jonathanhaigh-arm: Why are you using Python to get the value of `MBL_MIN_PASS_LEN` rather than just `${MBL_MIN_PASS_LEN}`?

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+
+MBL_MIN_PASS_LEN ?= "5"
```
jonathanhaigh-arm: Where does this `"5"` come from?

meta-mbl-distro/recipes-extended/shadow/shadow_%.bbappend
```diff
@@ -0,0 +1,29 @@
...
+
+MBL_MIN_PASS_LEN ?= "5"
+
+do_install_append () {
+
+    # We want to use SHA512 as encryption method to generate passwords
+    sed -i 's/^#ENCRYPT_METHOD.*$/ENCRYPT_METHOD SHA512/' ${D}${sysconfdir}/login.defs
+
+    # We want the minimum password length set by MBL_MIN_PASS_LEN
+    sed -i 's/^PASS_MIN_LEN.*$/PASS_MIN_LEN ${@d.getVar('MBL_MIN_PASS_LEN', True)}/' ${D}${sysconfdir}/login.defs
```
jonathanhaigh-arm: Why use Python rather than just `${MBL_MIN_PASS_LEN}`?

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
jonathanhaigh-arm: I'm confused. This code seems to imply that login.defs already contains a `PASS_MIN_LEN` setting, but we haven't added it in this PR. If the shadow packages already support a `PASS_MIN_LEN` setting then why aren't they already checking it in {user,group}{add,mod}? If they already have this setting then why is the upstreaming status "Inappropriate [OE specific]" - shouldn't they be using it (like your patch does)?

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
diego-sueiro: It is the default:
https://github.com/shadow-maint/shadow/blob/4.6/etc/login.defs#L205

jonathanhaigh-arm: Hmmm. I don't really like us repeating this in the recipe. We could just get rid of it and protect the `sed` command with:
```Shell
if [ -n "${MBL_MIN_PASS_LEN}" ]; then
```

I don't know whether it's worth it though... If you don't want to then can you add a comment explaining where the `"5"` comes from please?

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
diego-sueiro: The upstream doesn't support plain text password and it was added for oe-core via this patch:
https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-extended/shadow/shadow.inc#L28

In the upstream the `PASS_MIN_LEN` is only checked when using or `passwad` commands VIA `new_password` function call.

I see a typo in the `MBL_MIN_PASS_LEN`, as it should be `MBL_PASS_MIN_LEN`.

jonathanhaigh-arm: Ah, okay, that explains it... So should the patches go in a separate bbappend bbappend in oe-core-mbl and get upstreamed to oe-core?

jonathanhaigh-arm: FWIW, I'm a bit nervous about these patches anyway. If we can't upstream them then they seem like a pretty big maintenance burden.

diego-sueiro: I can move this patch to openembedded-core-mbl and we can try to upstream it.
I just like to mention that this patch is only applied for the native package, anyway.

jonathanhaigh-arm: > I just like to mention that this patch is only applied for the native package, anyway.

Huh? So what's the point then? We've already got checks at build time that the root password is long enough.

jonathanhaigh-arm: Why isn't the patch applied for the target package?

diego-sueiro: > Why isn't the patch applied for the target package?

Because in the target it is not supposed to support plain text passwords.

diego-sueiro: > > I just like to mention that this patch is only applied for the native package, anyway.
> 
> Huh? So what's the point then? We've already got checks at build time that the root password is long enough.

Ok, I'll remove it.

jonathanhaigh-arm: Oh, right, good :-)

I suggest we just get rid of the patches then. I don't think they're worth the maintenance burden.

meta-mbl-distro/classes/mbl-extrausers.bbclass
rwalton-arm: FWIW if you don't care about the exception's type and only care about passing an error message to the user you can just invoke `__str__`. Also worth noting is that `IOError` is an alias for `OSError` in python 3.3 and above so you might not need to catch the `IOError` explicitly. However, I'd agree with Jonathan that it's probably just better to let the exceptions propagate rather than trying to catch them here.

meta-mbl-distro/recipes-extended/shadow/shadow_%.bbappend
diego-sueiro: Addressed in 4a3571b.

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
diego-sueiro: Addressed in 4a3571b.

meta-mbl-distro/recipes-extended/shadow/shadow-sysroot_%.bbappend
diego-sueiro: Addressed in 4a3571b.

meta-mbl-distro/classes/mbl-extrausers.bbclass
diego-sueiro: Addressed in 4a3571b.

meta-mbl-distro/classes/mbl-extrausers.bbclass
diego-sueiro: Addressed in 4a3571b.

meta-mbl-distro/classes/mbl-extrausers.bbclass
jonathanhaigh-arm: Thanks. Resolving.

jonathanhaigh-arm:
> Implements: https://jira.arm.com/browse/IOTMBL-2374

When squashing and merging please add this to the commit message. (Also note that the part of the commit message about the patch is no longer relevant).

PR #674, opened 2019-08-20
Partitions: make use of all of a device's storage

jonathanhaigh-arm:
Extend the home partition to take up all of a device's remaining disk
space so that we don't waste it. The home partition was intentionally
made the last partition so that we could use extra space for apps (which
are stored on the home partition).

While I'm here, fix a bug in our imx8mmevk-mbl config - we currently
only support using SD Cards with this board rather than the on-board
eMMC so don't set the flash erase block size to the size used by the
eMMC.

For IOTMBL-697: Final sizing - handle free space

**Jenkins links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/143/ (succeeded).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/145/ (succeeded; after changes for review).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/146/ (in progress; after turning feature off by default).

**Lava links**
* http://e120856-lin/results.php?image_version=custom&image_number=143&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. Loads of timeouts.
* http://e120856-lin/results.php?image_version=custom&image_number=145&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job. Loads of timeouts.
 * http://e120856-lin/results.php?image_version=custom&image_number=146&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job (after turning off feature by default; everything passed).

**Note**
This change will make images a lot bigger (~7GiB for most boards) so flashing images to target boards with `dd` will take a lot longer. `bmaptool copy` users should be unaffected unless they use the `--nobmap` option.


meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -519,6 +561,13 @@ python __anonymous() {
             prev_size=prev_part_info.get("size_KiB"),
             prev_fs_part_number=prev_fs_part_number,
         )
+        if _process_bool_var(d, part_name, "FILL_STORAGE"):
+            part_info["size_KiB"], part_info["size_MiB"] = _extend_part_to_end_of_storage(
+                d,
+                part_name,
+                size_KiB=part_info["size_KiB"],
```
rwalton-arm: Is there a reason the `size_KiB` and `offset_KiB` are keyword args while the other two are positional args? It doesn't look like they have default values and it's no less clear what these parameters are than `d` or `part_name` given the way this function is invoked.

jonathanhaigh-arm: I suppose it was just because there are a bunch of functions in this chunk of code whose first two parameters are `d` and `part_name` (in that order) so I thought those two args would need less explanation that the other two.

I can see that the inconsistency is confusing though. I'll switch to just using positional args in this invocation.

jonathanhaigh-arm: Switched the keyword args to positional args in https://github.com/ARMmbed/meta-mbl/pull/674/commits/767abb3d03d247c01085888157438083a029b385.

meta-freescale-3rdparty-mbl/conf/machine/imx7d-pico-mbl.conf
```diff
@@ -126,3 +126,5 @@ MBL_WKS_BOOTLOADER2_IS_BL3 = "1"
 
 # As reported by "mmc extcsd read /dev/mmcblk1"
 MBL_FLASH_ERASE_BLOCK_SIZE_KiB = "512"
+
+MBL_WKS_STORAGE_SIZE_MiB = "7373"
```
rwalton-arm: Should we have a comment here explaining where the extra 205MiB comes from compared to most of the other boards? Maybe this is obvious but I don't understand why there's a difference.

diego-sueiro: I think we should have a comment about the storage size for all machines.

jonathanhaigh-arm: Yes, we should. My lack of comments was also hiding my sloppy process for coming up with the numbers...

jonathanhaigh-arm: Fixed in https://github.com/ARMmbed/meta-mbl/pull/674/commits/7041a0d1661c98432bf2efb4e4e2abaa7ec32222.

meta-freescale-3rdparty-mbl/conf/machine/imx6ul-des0258-mbl.conf
```diff
@@ -88,3 +88,6 @@ MBL_WKS_BOOTLOADER2_SIZE_KiB = "1024"
 
 # Guess based on value for imx6ul-pico
 MBL_FLASH_ERASE_BLOCK_SIZE_KiB = "512"
+
+# Guess based on value for imx6ul-pico
+MBL_WKS_STORAGE_SIZE_MiB = "3616"
```
diego-sueiro: @ARMmbed/team-embeddedplanet,
Can you please tell us the output of the following command in the des0258 board?
`blockdev --getsize64 /dev/mmcblk0`

jonathanhaigh-arm: I don't have access to a des0258. I suppose I can ask @bryanodonoghue.

jonathanhaigh-arm: Ah, I didn't notice you'd already tagged mentioned @ARMmbed/team-embeddedplanet.

jonathanhaigh-arm:
After discussing this change with @diegorusso, we've decided that I will turn this feature off by default because our test infrastructure can't really handle the large image sizes at the moment. I've raised https://jira.arm.com/browse/IOTMBL-2433 for dealing with our test infrastructure.

meta-freescale-3rdparty-mbl/conf/machine/imx6ul-des0258-mbl.conf
diego-sueiro: I'll resolve this conversation since we haven't received any response @ARMmbed/team-embeddedplanet.
If this value is different we can change it in another PR.

PR #676, opened 2019-08-21
distro:create-update-payload: Compress update files using gzip.

rwalton-arm:
The recently committed create-update-payload script was not compressing
the files within the tar archive correctly. So use gzip to compress the
files and pack them into an uncompressed tar archive.

On the target we use gunzip to decompress the files, this has been tested
on warp7.


meta-mbl-distro/scripts/create-update-payload
```diff
@@ -179,12 +180,23 @@ def create_payload_tar_archive(tar_archive_name, dir_path):
     :param tar_archive_name str: Path to the tar archive.
     :param dir_path Path: Path to the directory of the files to archive.
     """
+    files_for_tar = []
+    for fname in dir_path.iterdir():
+        gzip_fname = "{}.gz".format(
+            list(fname_var_map.keys())[
+                list(fname_var_map.values()).index(fname.name)
```
jonathanhaigh-arm: This confuses me... IUUC you're using the dict to find a key from a value. Would it be clearer if we just had another dictionary that maps in the opposite direction:
```
{ filename : varname for varname, filename in fname_var_map.items() }
```

Can we at least have a comment about what's going on here please?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -179,12 +180,23 @@ def create_payload_tar_archive(tar_archive_name, dir_path):
     :param tar_archive_name str: Path to the tar archive.
     :param dir_path Path: Path to the directory of the files to archive.
```
jonathanhaigh-arm: Can you document the new param please?

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: e5be650

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: e5be650

PR #688, opened 2019-08-23
Partitions: adjust sizes

jonathanhaigh-arm:
* Make all partition sizes multiples of 16MiB.
* Increase boot partition size to allow for future increases in FIT
  image size.
* Increase scratch partition size so that it can hold a full root
  partition image if it needs to.

IOTMBL-696: Partition clean up

**Jenkins links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/148/ (in progress).

**Lava links**
* coming soon.


meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -204,7 +204,7 @@ MBL_PARTITION_NAMES += "WKS_BOOTLOADER_FS"
 #
 MBL_BOOT_MOUNT_POINT = "${MBL_BOOT_DIR}"
 MBL_BOOT_LABEL ?= "boot"
-MBL_BOOT_DEFAULT_SIZE_MiB ?= "48"
+MBL_BOOT_DEFAULT_SIZE_MiB ?= "128"
```
rwalton-arm: Is there a reason we're increasing this almost 3x? Are we expecting the FIT image to grow that much larger? Is this within the typical size range for a boot partition on embedded systems? This is more out of curiosity than to gate the PR.

jonathanhaigh-arm: It's mainly because the FIT image on this partition contains the initramfs, and being able to have a large initramfs could be quite useful to some people.

jonathanhaigh-arm: > Is this within the typical size range for a boot partition on embedded systems?

I think this is on the larger side for embedded systems. If we add support for a targets with  smaller flash devices then we can still change the default sizes on a per-MACHINE basis.

PR #691, opened 2019-08-29
signing: Set TTL argument defaults.

rwalton-arm:
Set default values for the TTL and CA TTL arguments in the signctl CLI.


grandpaul:
Looks good to me. Have you asking jenkins to test this?

rwalton-arm:
Thanks for the review @grandpaul. This doesn't build through our build system yet, it's up to the user to manually install at the moment. I don't think there's any point in providing a Jenkins link as it won't actually be building this code.

PR #697, opened 2019-09-03
u-boot: production image with no delay.

grandpaul:
IOTMBL-2432: u-boot command line still available for production images

We set CONFIG_BOOTDELAY=-2 so that the u-boot does autoboot but
never break into the prompt. Thus fixes IOTMBL-2432.

Signed-off-by: Ying-Chun Liu (PaulLiu) <paulliu@debian.org>


grandpaul:
jenkins
http://jenkins.mbed-linux.arm.com/view/paulliu/job/paulliu-00-jenkins-test/319/

PR #699, opened 2019-09-03
linux-fslc: imx6ul: Fix xtest and reboot error on pico6

bryanodonoghue:
This patch switches to a pico6 defconfig that switches off the cpu idle
driver for the imx6ul.

One of the components that get downclocked as part of the idle process is
breaking runtime OP-TEE either on the down-clock or the restoral.

Switching off the imx6ul/sx CPU idle driver prevents the down clocking.

The ideal fix would determine which one of the various blocks being
quiesced is responsible for the error and isolate that component from the
idle process.

This change should be viewed as a temporary fix to that end.

Signed-off-by: Bryan O'Donoghue <bryan.odonoghue@linaro.org>


bryanodonoghue:
Having a bit of trouble with Jenkins but please find a simple update for Pico Linux kernel build that addresses

- xtest
- reboot

on Pico 6

meta-freescale-3rdparty-mbl/recipes-kernel/linux/linux-fslc_%.bbappend
```diff
@@ -8,12 +8,12 @@
...
 PV = "${LINUX_VERSION}+git${SRCPV}"
 LINUX_VERSION = "4.14.112"
-SRCREV = "937ee3dd236ed01a6be33fbabc79a528d62b1036"
+SRCREV = "46d7ce67b4e5bab34e42b4e857a7e0cbe9580998"
 SRCREV_imx6ul-des0258-mbl = "83d967e82fe285635be7845c6b578fee2c441996"
 
 KBUILD_DEFCONFIG_imx7s-warp-mbl ?= "warp7_mbl_defconfig"
 KBUILD_DEFCONFIG_imx7d-pico-mbl ?= "pico_mbl_mx6_mx7_defconfig"
-KBUILD_DEFCONFIG_imx6ul-pico-mbl ?= "pico_mbl_mx6_mx7_defconfig"
+KBUILD_DEFCONFIG_imx6ul-pico-mbl ?= "pico_mbl_mx6_mx7_no_cpuidle_defconfig"
```
diego-sueiro: As soon as this is a temporary fix, why we don't use the kernel `.cfg` file to disable the cpuidle?

bryanodonoghue: seems simpler in the time available to have a separate config - than figure out how to apply a .cfg in the preconfigure_append() of the imx6ul alone

diego-sueiro: But the merge-config.sh call is already in [place](https://github.com/ARMmbed/meta-mbl/blob/warrior-dev/meta-freescale-3rdparty-mbl/recipes-kernel/linux/linux-fslc_%25.bbappend#L44).

All you have to do is to add `SRC_URI_append_imx6ul-pico-mbl = "file://imx6ul-disable-cpu-idle-mbl.cfg"` in the `meta-freescale-3rdparty-mbl/recipes-kernel/linux/linux-fslc_%.bbappend` recipe.

bryanodonoghue:
http://jenkins.mbed-linux.arm.com/view/bod/job/bod-warrior-dev-00/19/

PR #704, opened 2019-09-05
Updates: Add boot partition updates to create-update-payload

rwalton-arm:
The update payload creation script now supports creating a payload for
the boot partitions and BLFS on raspberry pi 3.

Rework CLI to use 'bootloader components' and 'kernel' instead of
'bootloader slots' and 'boot partition', so users don't need to worry
about the details of the location of the components to update on the target.

Users now just specify either bootloader component 1,2 or kernel.

JIRA: IOTMBL-2284/2405


meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
+            src_path = path[0]
+            dst_name = path[1].name
+        else:
+            src_path = path
+            dst_name = path.name
+        # copy2 preserves file metadata and follows symlinks
+        shutil.copy2(str(src_path), str(dst_dir / dst_name))
+
+
+def copy_single_file_modifying_name(path, new_name, dst_dir):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: Actually, where this function is called... Couldn't we just use `copy_files_to_subdir` instead? That's just a more general version of this, albeit without the length check.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
+    tar.addfile(tar_info, buf)
+
+def create_tar_archive(tar_archive_name, dir_path, fname_var_map):
     """
-    Create the payload tar archive.
+    Create a tar archive.
 
     The tar archive is uncompressed but the files within it are compressed
-    using xz.
+    using gz.
```
jonathanhaigh-arm: Why switch from xz to gz? IIUC, xz tends to give much better compression at the cost of more computation on the compressing side, but not much more computation on the decompressing side, so it seems perfect for IoT.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
-        gzip_fname = "{}.gz".format(
-            list(fname_var_map.keys())[
-                list(fname_var_map.values()).index(fname.name)
-            ]
-        )
+        for var in fname_var_map:
+            if fname.is_dir():
+                gzip_fname = "{}.tar.gz".format(fname.name)
+            elif fname.name == var:
+                gzip_fname = "{}.tar.gz".format(var)
```
jonathanhaigh-arm: I really don't like these filenames... See my comments for https://github.com/ARMmbed/mbl-core/pull/208/.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
+        with tarfile.open(str(gzip_path), mode="w:gz") as tar_file:
+            tar_file.add(str(fname), arcname=gzip_path.name)
 
     with tarfile.open(tar_archive_name, mode="w") as tar:
+        add_payload_format_version_to_tar_archive(tar)
         for arc_path in files_for_tar:
             tar.add(str(arc_path), arcname=arc_path.name)
 
 
+def _bootloader_one_with_kernel(bootloader_components, tinfoil):
```
jonathanhaigh-arm: I don't think the name of this function is clear enough to make its single caller easier to read than if the code was inlined.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
+    return (
+        get_bitbake_conf_var("MBL_WKS_BOOTLOADER1_SKIP", tinfoil) == "1"
+        and "1" in bootloader_components
+    )
+
+
+def _kernel_with_bootloader_one(bootloader_components, tinfoil):
```
jonathanhaigh-arm: I don't think the name of this function is clear enough to make its single caller easier to read than if the code was inlined.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -97,128 +103,279 @@ def parse_args():
...
+    return get_bitbake_conf_var("MBL_WKS_BOOTLOADER1_SKIP", tinfoil) == "1"
+
+
+def get_payload_files_data(args, tinfoil):
```
jonathanhaigh-arm: I think this function needs a docstring.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -45,13 +44,15 @@ TOPDIR is ./mbl-manifest/build-mbl-development or
...
+import io
 import pathlib
 import re
 import shutil
 import sys
 import tarfile
 import tempfile
-
+import time
+import zipfile
```
jonathanhaigh-arm: I don't see where `zipfile` is used.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -72,23 +73,28 @@ bitbakepath = scriptpath.add_bitbake_lib_path()
...
-        "--bootloader-slots",
+        "--bootloader-components",
         nargs="+",
-        help="Space separated list of bootloader slot numbers. e.g 1 would"
-        " correspond to bootloader slot one.",
+        help="Space separated list of bootloader component numbers. e.g 1 would"
+        " correspond to bootloader component 1.",
+    )
+    parser.add_argument(
+        "-f",
```
jonathanhaigh-arm: Why is this `-f` rather than e.g. `-k`?

jonathanhaigh-arm:
@rwalton-arm is on holiday so I've fixed a couple of things that I've got time to do before I also go on holiday.

Due to pressure to get the next release out of the door, I don't think we'll be able to address all of the comments right now, so I've created https://jira.arm.com/browse/IOTMBL-2469 to address them after the release.

jonathanhaigh-arm:
@diego-sueiro, can you review please? I've added a bunch of comments but I propose leaving them unaddressed until after the next release. Please let me know where you don't think that's acceptable.

I'm in the process of doing tests on each of our supported boards...

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/704/commits/da9a2b3022d429c15a8bbe0ec277e2d24a98fbf9.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -72,153 +74,377 @@ bitbakepath = scriptpath.add_bitbake_lib_path()
...
         "-b",
-        "--bootloader-slots",
-        nargs="+",
-        help="Space separated list of bootloader slot numbers. e.g 1 would"
-        " correspond to bootloader slot one.",
+        "--bootloader-component",
+        dest="bootloader_components",
+        help="Add bootloader component N to the payload.",
+        metavar="N",
+        action="append",
```
rwalton-arm: I don't get why this was changed. Previously, because I was using `nargs="+"`, I could just give a space separated list of the bootloader slots, e.g `-b 1 2`, with this change we have to keep repeating the `-b` option e.g `-b 1 -b 2`. I'm changing this back to use `nargs="+"`

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: Removed unused imports in ac01602

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: Removed suffixes from actual files in ac01602

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: Changed to use `xz` in ac01602

meta-mbl-distro/scripts/create-update-payload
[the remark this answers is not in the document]
rwalton-arm: You removed this function in 9d9387832787dfb298c3b638ad1ecb4b11547c5c - resolving

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: I agree this check should be improved, but I'm leaving it as-is for now.

meta-mbl-distro/scripts/create-update-payload
rwalton-arm: I agree this check should be improved, but I'm leaving it as-is for now.

meta-mbl-distro/scripts/create-update-payload
diego-sueiro: addressed in 6a9f0c1923038a73e67474879819cb529a0df79a

PR #712, opened 2019-09-16
meta-mbl-distro:busybox: Move dd config to mbl-distro

rwalton-arm:
dd was only configured to have obs and ibs options on some platforms, as
the config file was located in meta-freescale-mbl. We want these options
on all platforms, so move the dd_ibs_obs.cfg to meta-mbl-distro and add
it to the SRC_URI in our busybox bbappend.

Jenkins: http://jenkins.mbed-linux.arm.com/job/rw-test/151/


meta-mbl-distro/recipes-core/busybox/busybox_%.bbappend
```diff
@@ -13,4 +13,4 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 # if_updown_ext_dhcp.cfg unsets CONFIG_FEATURE_IFUPDOWN_EXTERNAL_DHCP.
 # This means busybox ifup/ifdown will ignore an external managers such as ConnMan.
-SRC_URI += "file://if_updown_ext_dhcp.cfg"
+SRC_URI += "file://if_updown_ext_dhcp.cfg file://dd_ibs_obs.cfg"
```
diego-sueiro:
```suggestion
SRC_URI += "file://if_updown_ext_dhcp.cfg \
    file://dd_ibs_obs.cfg \
    "
```

PR #714, opened 2019-09-18
meta-freescale-mbl:u-boot-imx:imx8mmevk-mbl: Enable FEC without CMD_NET

diego-sueiro:
On imx8mmevk platform the FEC was only initialized when the
CONFIG_CMD_NET was enable.
For production images we don't enable the CMD_NET but still want the FEC
to be initialized and for that we also have to disabe the DM_ETH.


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/94/

PR #722, opened 2019-09-27
Fix up 0.8 documentation

jjohnson-arm:
* Release note was pointing to wrong tagged version
* meta-mbl docs were out of date with distribution info and BSP link


diego-sueiro:
Should this be backported to warrior-dev?

PR #724, opened 2019-09-30
meta-mbl-distro: Install supplied Authorized Keys and disable passwd login

diego-sueiro:
This patch introduces the mbl-external-ssh-pub-key class responsible
for adding the ssh authorized_keys file per user in the target rootfs.
The format of the SSH public key filename is ${username}_something.
The __username__ is mandatory in order to the installing function to
be able to know to which user home directory the public key should be
copied to (e.g.: /home/root/.ssh/authorized_keys).

In the case of building the 'mbl-production' DISTRO, if the
MBL_SSH_AUTHORIZED_KEYS_FILENAMES is/are not present the build will fail,
but for 'mbl-development' DISTRO image, only a warning will be displayed.

It also introduces the 'dis-passwd-login' dropbear PACKAGECONFIG to
disable ssh password login and is the default for production images.


diego-sueiro:
Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/98/
Sync: https://github.com/ARMmbed/mbl-tools/pull/312

PR #730, opened 2019-10-03
create-update-payload: Update for new payload format

jonathanhaigh-arm:
We're going to be using swupdate for part of our firmware update flow,
so change our payload format to something that swupdate understands.
I.e. a cpio archive containing a file for each component.

I've also changed the formats of some of the individual component
payloads:
* Payloads for bootloader components are no longer .tar.xz files. They
  just contain a single file so there's no need for the tar archive and
  they are now just plain .xz files.
* The payload for the rootfs component is now called ROOTFS.tar.xz to
  match the naming scheme of other components.
* Instead of having individual .ipk files in the overall payload
  archive, there's now an APPS.tar.xz file that contains all of the
  apps.

As well as making changes for the new payload format:
* I kept getting confused because sometimes lists of files to add to a
  payload contained tuples because the name of a file should be
  different in the payload archive to its name outside of the payload.
  I've introduced the new ArchivedFileSpec that holds the original path
  of a file and the name it should have in an archive. Now lists are
  heterogeneous and contain just ArchivedFileSpecs.
* rootfs and app payload support was added in a bit of a rush so the
  rootfs and apps were treated as special cases in a number of places.
  Refactor so that rootfs and apps are treated more similarly to
  bootloader and boot partition udpates.
* Add a "--verbose" option because sometimes I want to see which files
  are being added to the payload.

For IOTMBL-2540: Robust update: switch to using swupdate's payload
format

**Testing**
* Manual test of multi-component update of all components on imx7s-warp using mbl-firmware-update-manager.
* Manual test of multi-component update of all components on other platforms.
* Manual test of individual component updates on a platform.

**Lava links**
* http://e120856-lin/results.php?image_version=custom&image_number=162&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_job
  * All 6_rootfs-update-POST_CHECK tests failed, as expected - the rootfs update tests haven't been updated to use the new payload format yet.
  * There were Wi-Fi test fails on imx8mmevk (https://lava.mbedcloudtesting.com/scheduler/job/69300). I don't see how these could be related to the changes here. The job succeeded when re-submitted (https://lava.mbedcloudtesting.com/scheduler/job/69389).

**Jenkins links**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/160/
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/161/ (after _strict_path_resolve fix)
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/162/ (after `itertools.chain.from_iterable` change)

**Related PRs**
* mbl-core PR to update arm_update_activate.sh: https://github.com/ARMmbed/mbl-core/pull/214
* mbl-tools PR to update build-update-payloads.py: https://github.com/ARMmbed/mbl-tools/pull/334

**Notes**
* Need to update mbl-core SRCREV after mbl-core PR.


meta-mbl-distro/scripts/create-update-payload
```diff
@@ -239,108 +286,131 @@ def get_files_for_boot_part(deploy_dir, tinfoil=None):
...
-            shutil.copy2(str(src_path), str(dst_dir / dst_name))
-        except FileNotFoundError:
-            bb.fatal(
-                "File for the {} update component was not found at path {}. "
-                "Have you built the update components?".format(
-                    dst_name, str(src_path)
-                )
+    rootfs_filename = "{}-{}.tar.xz".format(image_name, machine)
+    try:
+        rootfs_image_path = pathlib.Path(deploy_dir, rootfs_filename).resolve()
```
rwalton-arm: We resolve this in the `ArchiveFileSpec` - probably don't need to resolve it here.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -239,108 +286,131 @@ def get_files_for_boot_part(deploy_dir, tinfoil=None):
...
 
-def copy_files_to_subdir(paths, dst_dir):
+def get_archived_file_spec_for_rootfs(image_name, tinfoil):
+    """
+    Get an ArchivedFileSpec for the root partition image.
     """
-    Copy files to a destination directory, renaming them when specified.
+    deploy_dir = get_bitbake_conf_var("DEPLOY_DIR_IMAGE", tinfoil)
+    if not deploy_dir:
+        bb.fatal("Failed to get value of DEPLOY_DIR_IMAGE variable")
```
rwalton-arm: I'm wondering if we should move this error checking to the `get_bitbake_conf_var` function? IIRC I think it was previously relying on some bitbake variables being `None` because of the way rootfs and app updates were plumbed in. Maybe we can just fail now if any bitbake variable is `None` - but I'm not sure.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -239,108 +286,131 @@ def get_files_for_boot_part(deploy_dir, tinfoil=None):
...
+    except FileNotFoundError:
+        bb.fatal(
+            "The rootfs image with filename {} was not found. Have you "
+            "provided the correct image-name? Have you built the rootfs "
+            "image and is it located in {}?".format(
+                rootfs_filename, deploy_dir
             )
+        )
 
+    if not rootfs_image_path.exists():
```
rwalton-arm: Oh, we're creating the resolved `Path` because we check if the file exists here? Maybe we should actually move this check into the `ArchivedFileSpec::__init___` method? It would make the calling code cleaner and it's probably good to know if any of these archive files don't exist.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -74,7 +67,48 @@ import argparse_oe
...
+            resolution).
+        """
+        self.path = pathlib.Path(path)
+        if archived_path is None:
+            self.archived_path = path.name
+        else:
+            self.archived_path = pathlib.Path(archived_path)
+
+        # We don't want symlinks in our payloads
+        self.path = self.path.resolve()
```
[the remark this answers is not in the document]
jonathanhaigh-arm: That won't work when `path` is a symlink - if `archived_path` isn't given then it should be the basename of the `path` before it's resolved.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -74,7 +67,48 @@ import argparse_oe
...
+
+
+def print_verbose(s):
+    """Print only if in verbose mode"""
```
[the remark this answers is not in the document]
jonathanhaigh-arm: Yeah, I considered that and then chose not to use it because we're already using BitBake's functions for loging fatal errors.

I wouldn't mind getting rid of the BitBake style error reporting just using a logger + exit codes or uncaught exceptions.

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -225,12 +272,12 @@ def get_files_for_boot_part(deploy_dir, tinfoil=None):
     if not image_boot_files:
         bb.fatal("Failed to find any files in IMAGE_BOOT_FILES")
 
-    boot_file_specs = image_boot_files.split()
+    boot_file_entries = image_boot_files.split()
     boot_files = list(
         itertools.chain(
             *[
```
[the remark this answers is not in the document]
jonathanhaigh-arm: I don't understand what your suggesting. How could I join up all the lists in the list comp?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -239,108 +286,131 @@ def get_files_for_boot_part(deploy_dir, tinfoil=None):
...
+                'Adding "{}" to "{}" as "{}"'.format(
+                    file_spec.path, file_name, file_spec.archived_path
+                )
+            )
+            tar.add(str(file_spec.path), arcname=str(file_spec.archived_path))
+    print_verbose('Adding "{}" to payload'.format(file_name))
 
-        copy_files_to_subdir(paths, dst_dir)
 
+def read_chunks(f):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: This is so that we don't read the whole file into memory at once. I was hoping there'd be a simpler way to do this, but I can't see how I can leverage the buffering in FileIO (or any of the other classes in the io module) to do this.

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Yeah, good point. I'll go over the file existence checks and resolution and see what I can do...

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Good point. I'll take a look.

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Done in https://github.com/ARMmbed/meta-mbl/pull/730/commits/550dd85fe34f39728ae8110a922ff15d5eadc2a0.

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/730/commits/550dd85fe34f39728ae8110a922ff15d5eadc2a0.

meta-mbl-distro/scripts/create-update-payload
jonathanhaigh-arm: Factored out in https://github.com/ARMmbed/meta-mbl/pull/730/commits/7613c6bf408aa394e97ff010094b7102dbb46ae1.

I had to keep an option to not fail when the variable isn't set for when we're checking the `_SKIP` variables. While I was in that code, I also factored out the code to check whether a partition is skipped.

jonathanhaigh-arm:
@rwalton-arm, I've added https://github.com/ARMmbed/meta-mbl/pull/730/commits/5c4913741857123d9285158c45e1e6f390a5231c since you reviewed. Could you take a look please?

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -74,7 +68,72 @@ import argparse_oe
...
+    # In python < 3.6, pathlib.Path.resolve() doesn't have a "strict"
+    # parameter, but it always does strict resolution so we just have to avoid
+    # passing the "strict" parameter.
+    resolve_args = {}
+    if "strict" in inspect.getfullargspec(path.resolve).args:
+        # In python >= 3.6, pathlib.Path.resolve() has a "strict"
+        # parameter, and we must set it to True to avoid the default
+        # value of False.
+        resolve_args = { "strict": True }
+    return pathlib.Path(path).resolve(*resolve_args)
```
rwalton-arm: Does this work? I would've thought you'd need to use dictionary unpacking for this keyword argument i.e `(**resolve_args)` looks like you're just using list unpacking here so it'll just pass the key in.

jonathanhaigh-arm: I tested it, and it did work... But then with list unpacking `"strict"` will be interpreted as the value for the `strict` parameter, and `"strict"` is truthy, so I suppose it would work by accident.

Will fix.

jonathanhaigh-arm: But then why do I bother doing that at all when I could just do the following?
```Python
def _strict_path_resolve(path):
    """
    Acts like "pathlib.Path(path).resolve(strict=True)" from Python >= 3.6.
    """
    if "strict" in inspect.getfullargspec(path.resolve).args:
        # In python >= 3.6, pathlib.Path.resolve() has a "strict"
        # parameter, and we must set it to True to avoid the default
        # value of False.
        return pathlib.Path(path).resolve(strict=True)
    # In python < 3.6, pathlib.Path.resolve() doesn't have a "strict"
    # parameter, but it always does strict resolution.
    return pathlib.Path(path).resolve()
```

jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/730/commits/05190574fcf05e6b6d49b499666f85d53c147410

meta-mbl-distro/scripts/create-update-payload
```diff
@@ -206,31 +284,26 @@ def get_path_for_bootloader_slot(bootloader_slot_name, deploy_dir, tinfoil):
...
-    boot_file_specs = image_boot_files.split()
+    boot_file_entries = get_bitbake_conf_var(
+        "IMAGE_BOOT_FILES", tinfoil
+    ).split()
     boot_files = list(
         itertools.chain(
             *[
-                img_boot_files_val_to_map_val(elem, deploy_dir)
-                for elem in boot_file_specs
+                img_boot_files_val_to_archived_file_specs(elem, deploy_dir)
```
rwalton-arm: I know I already approved this, but I'm wondering if we can do something like 
`list(itertools.chain.from_iterable(img_boot_files_val_to_archived_file_specs(elem, deploy_dir) for elem in boot_file_entries))` and save building the extra list in the comprehension?

jonathanhaigh-arm: Yeah, that is a bit neater.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/730/commits/eacb96ea2f1ec15419d9f415969e7058940bc456

jonathanhaigh-arm:
@rwalton-arm, I've update the mbl-core SRCREV. Can you re-approve please?

PR #728, opened 2019-10-01
signing: Add NXP signing support

rwalton-arm:
This patch adds support for signing imx6, 7 and 8 FIP images and BL2.

The BL2 image must have a root-of-trust public key hash patched into it.
This hash is used to verify the initial stages of the TF certificate
chain of trust.

This patch also refactors the CLI so the commands and options make more
sense.

Raspberry PI remains a special case, as BL2 is combined with BL1 in the
VC4 firmware binary, BL2 is also wrapped in its own FIP. So if users want
to re-sign BL2 on RPI3 they must use the rpi-vc4-fw option and not the BL2
 option.

PR #731, opened 2019-10-03
signing: Add NXP signing support (#728)

jjohnson-arm:
This patch adds support for signing imx6, 7 and 8 FIP images and BL2.

The BL2 image must have a root-of-trust public key hash patched into it.
This hash is used to verify the initial stages of the TF certificate
chain of trust.

This patch also refactors the CLI so the commands and options make more
sense.

Raspberry PI remains a special case, as BL2 is combined with BL1 in the
VC4 firmware binary, BL2 is also wrapped in its own FIP. So if users want
to re-sign BL2 on RPI3 they must use the rpi-vc4-fw option and not the BL2
 option.


jjohnson-arm:
Cherry-pick 9b1fbae from warrior-dev

PR #738, opened 2019-10-09
meta-mbl-distro:mbl-partitions.bbclass: fstab fs_passno defaults to 2

diego-sueiro:
To prevent the system from entering the emergency mode, because of
filesystem corruptions, we need to enable the fsck for all partitions.
This is accomplished by setting the fs_passno parameter greater than 0
in the /etc/fstab file. By default the "/" partition is 1 and all the
others are 2.


diego-sueiro:
**NOTE:** needs to be ported to mbl-os-0.9 branch.

Jenkins: http://jenkins.mbed-linux.arm.com/job/ds-warrior-dev/104/

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -46,7 +46,7 @@ MBL_DEFAULT_FS_FREQ ?= "0"
 
 # Default fs_passno (the sixth fstab field - see "man 5 fstab") for file system
 # partitions when MBL_<partition>_FS_PASSNO (documented below) isn't set.
-MBL_DEFAULT_FS_PASSNO ?= "0"
+MBL_DEFAULT_FS_PASSNO ?= "2"
```
diegorusso: Please, update the comment above.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -46,7 +46,7 @@ MBL_DEFAULT_FS_FREQ ?= "0"
 
 # Default fs_passno (the sixth fstab field - see "man 5 fstab") for file system
 # partitions when MBL_<partition>_FS_PASSNO (documented below) isn't set.
```
diego-sueiro:
```suggestion
# partitions when MBL_<partition>_FS_PASSNO (documented below) is set to 2 to
# force the fsck to run.
```

meta-mbl-distro/classes/mbl-partitions.bbclass
diego-sueiro: done.

PR #757, opened 2019-10-24
meta-mbl-distro:recipes-support: Add swupdate package

rwalton-arm:
We're using swupdate for robust update. This patch adds swupdate to
the packagegroup so it is included in the image.

We also need to configure swupdate so it doesn't depend on u-boot or
libubootenv as we don't have a u-boot environment. This patch also
adds a defconfig which sets CONFIG_BOOTLOADER_NONE and unsets
CONFIG_UBOOT so swupdate doesn't try to write to the
u-boot environment.

Jenkins: http://jenkins.mbed-linux.arm.com/job/rw-test/168/

LAVA: http://e120856-lin.cambridge.arm.com/results.php?image_version=custom&image_number=168&custom_version=rw-test&lava_query=&lava_owner=&failures=false&submitter=robwal02&sort=by_job

Merge after related PRs.

Related PRs:
https://github.com/ARMmbed/mbl-config/pull/63
https://github.com/ARMmbed/mbl-manifest/pull/132


meta-mbl-distro/recipes-core/packagegroups/packagegroup-mbl-production.bb
```diff
@@ -56,6 +56,7 @@ PACKAGEGROUP_MBL_PRODUCTION_PKGS_append = " connman"
 PACKAGEGROUP_MBL_PRODUCTION_PKGS_append = " connman-client"
 PACKAGEGROUP_MBL_PRODUCTION_PKGS_append = " systemd-net-conf"
 PACKAGEGROUP_MBL_PRODUCTION_PKGS_append = " systemd-watchdog-conf"
+PACKAGEGROUP_MBL_PRODUCTION_PKGS_append = " swupdate"
```
jonathanhaigh-arm: I think this is okay for now, but when we integrate swupdate with our firmware update solution, we should remove this and just make swupdate an rdependency of our `mbl-cloud-client` recipe (or our `updated` recipe once we have one).

meta-mbl-distro/recipes-support/swupdate/files/defconfig
```diff
@@ -0,0 +1,98 @@
...
+#
+# Swupdate Settings
+#
+
+#
+# General Configuration
+#
+CONFIG_SCRIPTS=y
+CONFIG_HW_COMPATIBILITY=y
+CONFIG_HW_COMPATIBILITY_FILE="/etc/hwrevision"
```
jonathanhaigh-arm: Does `/etc/hwrevision` get created at build time? If so, we might be able to use the hw compatibility feature, but I'd suggest that the file belongs in our factory config partition.

meta-mbl-distro/recipes-support/swupdate/files/defconfig
```diff
@@ -0,0 +1,98 @@
...
+CONFIG_SW_VERSIONS_FILE="/etc/sw-versions"
```
jonathanhaigh-arm: I'm not convinced swupdate's sw compatibility feature will be suitable for us... Can we turn it off? If we have to keep this file then it probably should live in our user config partition.

meta-mbl-distro/recipes-support/swupdate/files/defconfig
rwalton-arm: I'm testing with this config removed now. I had to manually create `/etc/hwrevision` in my previous test.

meta-mbl-distro/recipes-support/swupdate/files/defconfig
rwalton-arm: Testing it works with this switched off now.

rwalton-arm: I don't think we can actually turn this off, we can just specify the file it uses. If the file doesn't exist it looks like swupdate's `artifact_versions.c` creates it at `/etc/swversions` https://github.com/sbabic/swupdate/blob/f46e9e1ff4617aa876c9ede964615a8e806bca83/corelib/artifacts_versions.c

I'll try moving it to the user config partition.

meta-mbl-distro/recipes-support/swupdate/files/defconfig
rwalton-arm: Seems to work with the HW_COMPATIBILITY switched off in my simple test. I can't find much in the `swupdate` documentation about this config. Do you think we should leave it switched off?

jonathanhaigh-arm: I can only see it used in one place in the code (parser/parser.c). It looks like if we switch it off it will just ignore hardware compatibility checks in sw-description files.

I think we should turn it off until we decide we want to use it.

meta-mbl-distro/recipes-support/swupdate/files/defconfig
jonathanhaigh-arm: Mmm, okay. I suppose we'll just have to live with it in the user config partition.

rwalton-arm: I've had to hard-code the /user/config path in the `defconfig`. I tried to use `mbl-var-placeholders` but that wouldn't work as the task was running after the `swupdate` compilation task so the changes weren't being picked up.

jonathanhaigh-arm: Would something like the following work?
```
do_configure_prepend() {
    printf 'CONFIG_SW_VERSIONS_FILE="%s"' "${MBL_CONFIG_DIR}/sw-versions" >> "${WORKDIR}/defconfig"
}
```

rwalton-arm: Hmm... yeah... I should've thought of that! I'll try it.

PR #761, opened 2019-10-25
swupdate: Disable systemd integration.

rwalton-arm:
We don't want to use swupdate as a systemd service.
Disable the systemd integration by disabling CONFIG_SYSTEMD and
removing the systemd.service files the swupdate recipe installs.

Jenkins: http://jenkins.mbed-linux.arm.com/job/rw-test/171/
LAVA: http://e120856-lin.cambridge.arm.com/results.php?image_version=custom&image_number=171&custom_version=rw-test&lava_query=&lava_owner=&failures=false&submitter=robwal02&sort=by_device

There were 2 odd failures in the systemd test on imx7d-pico, one of them looked like a bmaptool issue and the other was the logrotate service failing to load. I resubmitted the job and the tests then passed. There was also a bthelper service failure on raspberrypi3 which is a known issue.


meta-mbl-distro/recipes-support/swupdate/swupdate_%.bbappend
```diff
@@ -2,11 +2,21 @@
...
+SYSTEMD_SERVICE_${PN} = ""
 
+# Change the location of the software versions file to our user config partition.
 do_configure_prepend() {
     printf 'CONFIG_SW_VERSIONS_FILE="%s"' "${MBL_CONFIG_DIR}/sw-versions" >> "${WORKDIR}/defconfig"
 }
+
+# Override to remove the installation of all systemd.service files. We need this as swupdate's systemd
+# integration is baked in to swupdate.inc in meta-swupdate.
+do_install() {
```
diego-sueiro: You can set `SYSTEMD_AUTO_ENABLE = "disable"`.
https://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#var-SYSTEMD_AUTO_ENABLE

jonathanhaigh-arm: TBH, I don't think that would be better - we don't want the service at all, whether it's started by default or not.

PR #763, opened 2019-10-29
create-update-manifest: add sw-description and testinfo

jonathanhaigh-arm:
* Add a sw-description file to our update payloads to tell swupdate how to
  install them.

* Add an option to create a "testinfo" file for an update payload that
  provides our test infrastructure with the information required to test
  an update using the payload.

* The script is getting a bit more complex, so to factorize it a bit
  more by:
  * Introducing "image" classes for each type of component image.
  * Introducing an UpdatePayload class to build a payload using the
    component images.
  * Splitting the code into modules.

**Notes**:
* The code here will not work until our update scripts can handle the new payload type. That will come later, together with testing.
* This change gives `create-update-payload` a new dependency on the `arpy` and `libconf` modules. These will need to be added to our build container image.

**Jenkins links**:
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/167/ (build "succeeded" but payloads weren't created. I've raised IOTMBL-2605 because the build should have failed).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/169/

**Related PRs (synchronize merges)**
* https://github.com/ARMmbed/mbl-core/pull/233 (changes on-target update scripts to accept new payload format)
* https://github.com/ARMmbed/mbl-tools/pull/355 (add new dependencies to build container and create testinfo files)


meta-mbl-distro/scripts/lib/update/payload.py
```diff
@@ -0,0 +1,190 @@
...
+import tempfile
+
+from update.util import get_bitbake_conf_var
+import update.images as images
+import update.swdesc as swdesc
+import update.testinfo as testinfo
+
+CURRENT_PAYLOAD_FORMAT_VERSION = 3
+
+PAYLOAD_FORMAT_VERSION_INFO = {
```
rwalton-arm: I'm not sure I like this apparent "constant" returning a dictionary. Could we make this a function call  instead? Python doesn't have  "const" or anything, but the convention is that these uppercase variables are "constant" values so it feels a bit misleading the make this a dictionary.

meta-mbl-distro/scripts/lib/update/testinfo.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import hashlib
```
rwalton-arm: TBH I don't think this test info stuff belongs in the `create-update-payloads` it feels like we're encoding too much knowledge of various tests we're going to run in our test farm in what is supposed to be a user-facing tool for generating update payloads. Could we not just have a separate script that generates this "testinfo" stuff? Things like the sha256 of the image can actually be used by `swupdate` for verification so maybe that belongs here.

meta-mbl-distro/scripts/lib/update/payload.py
jonathanhaigh-arm: But it's a constant dictionary. I'm trying to convey that the dictionary won't change after it has been constructed.

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        for line in control_file.read().decode("utf-8").splitlines():
+            field, value = line.split(sep=":", maxsplit=1)
+            if field.strip() == "Package":
+                return value.strip()
+        bb.fatal(
+            'Failed find app name in "control" file from "{}"'.format(ipk_path)
+        )
+
+
+class BootImage:
```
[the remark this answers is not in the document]
jonathanhaigh-arm: Yeah, that's a good idea.

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+    def generate_testinfo(self):
+        return [testinfo.file_compare("/proc/version")] + [
+            testinfo.file_sha256(
+                afs.path, pathlib.Path("/boot", afs.archived_path)
+            )
+            for afs in self._boot_files
+        ]
+
+    @staticmethod
+    def _get_archived_file_specs(deploy_dir, tinfoil):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: > I guess you could argue this method isn't changing the object's state so it should be static (or a free function)

Exactly - the method doesn't access any object state. Marking it as static means that readers immediately know that, so understanding the class, and the method itself, should be easier.

>  but I'm still not sure I like this shape.

Why?

meta-mbl-distro/scripts/lib/update/images.py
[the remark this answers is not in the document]
jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/763/commits/93b30b4fd29e29a5a699e5838e5d6ebfbedb7dab

meta-mbl-distro/scripts/lib/update/payload.py
rwalton-arm: I'm not sure why this nested dictionary is required anyway. What does having this "image_format_versions" key buy us rather than having a tuple of "image format versions"? Are we planning on extending the inner dictionary with more keys? I kind of see the need for the outer dictionary if we're planning on handling multiple payload format versions, but I don't get why we need the inner dict.

meta-mbl-distro/scripts/lib/update/images.py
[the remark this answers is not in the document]
rwalton-arm: I think I'd prefer we had another module with free functions rather than all these `staticmethods`; but as I said in the RL discussion, I'm not going to gate the PR on it, as it's my stylistic preference rather than having any technical argument for changing it.

meta-mbl-distro/scripts/lib/update/payload.py
jonathanhaigh-arm: I think an example will illustrate what I'm thinking best. Suppose we introduce a new format for rootfs images (e.g. an ext4 image file rather than a tar). Then I'd expect to update `PAYLOAD_FORMAT_VERSION _INFO` to look like this:
```Python
PAYLOAD_FORMAT_VERSION_INFO = {
    3: {
        "image_format_versions": {
            images.MBL_WKS_BOOTLOADER1_ID: 3,
            images.MBL_WKS_BOOTLOADER2_ID: 3,
            images.MBL_BOOT_ID: 3,
            images.MBL_ROOTFS_ID: 3,
            images.MBL_APPS_ID: 3,
        }
    },
    4: {
        "image_format_versions": {
            images.MBL_WKS_BOOTLOADER1_ID: 3,
            images.MBL_WKS_BOOTLOADER2_ID: 3,
            images.MBL_BOOT_ID: 3,
            images.MBL_ROOTFS_ID: 4,
            images.MBL_APPS_ID: 3,
        }
    }
}
```
 Payload format version 4 would have the same image format versions as payload version 3, except for the rootfs image format version which would be 4.

I think I made a mistake in adding this dict  before we actually have a mechanism for choosing different payload format versions though. It doesn't really make much sense on its own. I'll remove it for now.

jonathanhaigh-arm: Oh, I think I misunderstood your question... Either way, I think it will be addressed by me removing the dict.

jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/763/commits/174d5ce546fed12c21052193ef8922b975c4d28e.

meta-mbl-distro/scripts/lib/update/images.py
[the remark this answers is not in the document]
jonathanhaigh-arm: Addressed in https://github.com/ARMmbed/meta-mbl/pull/763/commits/ecb7eeeaacee2e09b5f38ef595037176c4fe2b5a

meta-mbl-distro/scripts/lib/update/testinfo.py
jonathanhaigh-arm: I've been thinking about this for a few days, and I've spoken to @jjohnson-arm about it (I meant to speak to you and @diegorusso about it too but I forgot that you'd be on holiday).

My position is this:
Emitting some info to aid testing is very useful, and generating it in a separate script would have issues:
  * If we used a separate script that used the BitBake environment then there's some chance that the two scripts will disagree about what's in the payload. 
  * If we had a separate script that interrogated an actual payload to work out what's in it then that script would have to deal with unpacking a payload, which would be a bunch of unnecessary effort.

So I definitely think we should have this script creating some information to help with testing. 

In principle, I do think the test info that's created is too prescriptive about what testing should be done - it should be providing information about the payload that will help with testing rather than specifying the tests that should run and IIUC, you, @jjohnson-arm and I agree on this.

I'm not convinced that at this point it would be pragmatic to change the approach though - it would make the testing code more complicated and I don't think it would be worth it.

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        rootfs_filename = "{}-{}.tar.xz".format(image_name, machine)
+        self._archived_file_spec = ArchivedFileSpec(
+            deploy_dir / rootfs_filename, "{}.tar.xz".format(MBL_ROOTFS_ID)
+        )
+
+        self.image_type = MBL_ROOTFS_ID
+        self.image_format_version = 3
+        self.archived_path = self._archived_file_spec.archived_path
+
+    def stage(self, staging_dir):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        _stage_single_file(staging_dir, self._archived_file_spec)
+
+    def generate_testinfo(self):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        return [
+            testinfo.file_timestamp_compare("/etc/build"),
+            testinfo.mounted_bank_compare("/"),
+        ]
+
+
+class WksBootloaderSlotImage:
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+    def __init__(self, bootloader_slot_name, deploy_dir, tinfoil):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        filename = get_bitbake_conf_var(filename_var_name, tinfoil)
+        self._archived_file_spec = ArchivedFileSpec(
+            deploy_dir / filename, "{}.xz".format(bootloader_slot_name)
+        )
+
+        self.image_type = bootloader_slot_name
+        self.image_format_version = 3
+        self.archived_path = self._archived_file_spec.archived_path
+
+    def stage(self, staging_dir):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        _stage_single_file(staging_dir, self._archived_file_spec)
+
+    def generate_testinfo(self):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        return [
+            testinfo.partition_sha256(
+                self.image_type, self._archived_file_spec.path
+            )
+        ]
+
+
+class AppsImage:
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+    def __init__(self, app_paths):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        self._apps = [ArchivedFileSpec(app) for app in app_paths]
+        self._validate_app_paths(self._apps)
+        self.image_type = MBL_APPS_ID
+        self.image_format_version = 3
+        self.archived_path = "{}.tar.xz".format(MBL_APPS_ID)
+
+    def stage(self, staging_dir):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        _stage_multi_file_component(
+            staging_dir, self.archived_path, self._apps
+        )
+
+    def generate_testinfo(self):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        return [
+            testinfo.app_bank_compare(self._get_app_name(afs.path))
+            for afs in self._apps
+        ]
+
+    @staticmethod
+    def _validate_app_paths(apps):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+            field, value = line.split(sep=":", maxsplit=1)
+            if field.strip() == "Package":
+                return value.strip()
+        bb.fatal(
+            'Failed find app name in "control" file from "{}"'.format(ipk_path)
+        )
+
+
+class BootImage:
+    def __init__(self, deploy_dir, tinfoil):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        self._boot_files = self._get_archived_file_specs(deploy_dir, tinfoil)
+        self.image_type = MBL_BOOT_ID
+        self.image_format_version = 3
+        self.archived_path = pathlib.Path("{}.tar.xz".format(MBL_BOOT_ID))
+
+    def stage(self, staging_dir):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+        _stage_multi_file_component(
+            staging_dir, self.archived_path, self._boot_files
+        )
+
+    def generate_testinfo(self):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/util.py
```diff
@@ -0,0 +1,90 @@
...
+        except FileNotFoundError:
+            bb.fatal(
+                'The file "{}" was not found.'
+                " Have you built all of the required components?".format(
+                    self.path
+                )
+            )
+
+    @staticmethod
+    def _strict_path_resolve(path):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: > Should we make this a 'public' free function? I think we could re-use it in other scripts, and it wouldn't make much sense to have to refer to the ArchivedFileSpec if we wanted to use it in a different context.

Agree.

> Thinking about it, the read_chunks generator and get_bitbake_conf_var would be also pretty useful for other scripts.

I agree about `read_chunks`. I'm not sure I agree about `get_bitbake_conf_var`... It does the same thing as BitBake's `getVar` but with error handling that is less generally useful.

Let me see what I can do...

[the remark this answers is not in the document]
jonathanhaigh-arm: I've moved `strict_path_resolve` and `read_chunks` into a fileutils lib. I've moved `get_bitbake_conf_var` into a tinfoilutils lib - I think it might be useful for other scripts that use Tinfoil, but not other BitBake stuff.

https://github.com/ARMmbed/meta-mbl/pull/763/commits/7553785672934548c33b8c6fda45f478b2855208

meta-mbl-distro/scripts/lib/update/images.py
```diff
@@ -0,0 +1,331 @@
...
+
+MBL_WKS_BOOTLOADER1_ID = "WKS_BOOTLOADER1"
+MBL_WKS_BOOTLOADER2_ID = "WKS_BOOTLOADER2"
+MBL_BOOT_ID = "BOOT"
+MBL_ROOTFS_ID = "ROOTFS"
+MBL_APPS_ID = "APPS"
+
+
+class RootfsImage:
+    def __init__(self, image_name, deploy_dir, tinfoil):
```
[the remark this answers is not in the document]
jonathanhaigh-arm: 7553785

meta-mbl-distro/scripts/lib/update/testinfo.py
rwalton-arm: My suggestion would've been to use the BitBake environment and write a new script, which calls the library code you've created, to find the payload images and create the test info from them. I think that would've lowered the probabilty of the two scripts disagreeing over the payload content? It would also remove the test info generation from the `create-update-payload` interface, without having to re-implement the logic from the `create-update-payload` script.

I won't gate the PR on this as it probably needs some more thought, but I do still think we should look into splitting out the test info from the `create-update-payload` script at some point.

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""PayloadImage subclass for apps images."""
+
+import arpy
```
rwalton-arm: I noticed we are just going to install this and `libconf` in our docker container. Even so I would prefer we provided a `requirements.txt` for users who don't use the mbl-tools scripts.

jonathanhaigh-arm: Okay.

jonathanhaigh-arm: Added requirements.txt in https://github.com/ARMmbed/meta-mbl/pull/763/commits/efb13e2bdcf34f162330be0acadc579fe1b30b37

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @property
+    def image_type(self):
+        """Implement method from PayloadImage ABC."""
+        return MBL_APPS_ID
+
+    @property
+    def image_format_version(self):
+        """Implement method from PayloadImage ABC."""
+        return 3
```
diego-sueiro: Can't we use a global variable to store the version number?

meta-mbl-distro/scripts/lib/mbl/update/bootimage.py
```diff
@@ -0,0 +1,122 @@
...
+
+    @property
+    def image_type(self):
+        """Implement method from PayloadImage ABC."""
+        return MBL_BOOT_ID
+
+    @property
+    def image_format_version(self):
+        """Implement method from PayloadImage ABC."""
+        return 3
```
diego-sueiro: Can't we use a global variable to store the version number?

meta-mbl-distro/scripts/lib/mbl/update/rootfsimage.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def image_type(self):
+        """Implement method from PayloadImage ABC."""
+        return MBL_ROOTFS_ID
+
+    @property
+    def image_format_version(self):
+        """Implement method from PayloadImage ABC."""
+        return 3
```
diego-sueiro: Can't we use a global variable to store the version number?

meta-mbl-distro/scripts/lib/mbl/update/wksbootloaderslotimage.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @property
+    def image_type(self):
+        """Implement method from PayloadImage ABC."""
+        return self._bootloader_slot_name
+
+    @property
+    def image_format_version(self):
+        """Implement method from PayloadImage ABC."""
+        return 3
```
diego-sueiro: Can't we use a global variable to store the version number?

meta-mbl-distro/scripts/lib/mbl/update/payload.py
```diff
@@ -0,0 +1,154 @@
...
+            _append_to_payload(
+                staging_dir_path, swdesc_name, output_path, create=True
+            )
+            for image in self.images:
+                image.stage(staging_dir_path)
+                _append_to_payload(
+                    staging_dir_path, image.archived_path, output_path
+                )
+
+    def create_testinfo_file(self, output_path):
```
diego-sueiro: I suppose that the testinfo will be added to all payloads created. In the production environment will we want this information? If not, I suggest we make this optional, unless this info can be potentially used to self validate the update without any additional host tools/tests.

meta-mbl-distro/scripts/lib/mbl/util/fileutil.py
```diff
@@ -0,0 +1,36 @@
...
+        # parameter, and we must set it to True to avoid the default
+        # value of False.
+        return pathlib.Path(path).resolve(strict=True)
+    # In python < 3.6, pathlib.Path.resolve() doesn't have a "strict"
+    # parameter, but it always does strict resolution.
+    return pathlib.Path(path).resolve()
+
+
+def read_chunks(f):
+    """Read 4KiB chunks from a file-like object."""
```
diego-sueiro: Any specific reason why we are using 4kiB? Is this machine dependant? If so I suggest we add a parameter for the chunk size and defaults to `4096`, if not I suggest renaming the function do `read_4kiB_chunks`.

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
rwalton-arm: IIUC we can have a different version number for each image, so a global variable sounds like the wrong choice for this.

meta-mbl-distro/scripts/lib/mbl/update/payload.py
rwalton-arm: There's a command line option for adding the test info.

diego-sueiro: Thanks for clarifying.

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
diego-sueiro: We can have a global variable per component which at the moment have the same value. I don't like the idea of returning a fixed number in a function.

rwalton-arm: Maybe we could have a class attribute in the base class and override it for each property in the concrete classes. However, I'm not sure it's worth the effort. These properties don't really behave like functions when called; they're intended to provide a layer of encapsulation for a class attribute you want as part of the public interface, so you can refactor or change the underlying class attribute without changing the public interface of the class. It's a little unusual to just return a fixed number, but python devs wouldn't find it too surprising as properties generally do just return "a value".
At the call site this would look something like `image.image_format_version == 3` so behaving more like an attribute than a function call.

jonathanhaigh-arm: I agree with Rob. When we start having multiple versions of each `Image` class then it would make less sense to use globals.

jonathanhaigh-arm: FWIW, when we bump versions, we'll be bumping the versions of each `Installer` class independently, and a single version that all the classes use won't work for that.

Before we do the next release, we'll need to support creating multiple versions of payloads within the script so that we can support updating from earlier versions of software over the air. At that point we'll have multiple versions of each image class so defining the version for a class outside the class itself will make less sense.

When we support multiple version of payload within the script then we'll need a richer mechanism for marking versions of classes and choosing the right one to use. I suggest we re-evaluate when implementing that mechanism.

meta-mbl-distro/scripts/lib/mbl/util/fileutil.py
jonathanhaigh-arm: I'll add a chunk size parameter. I don't really want the user to have to care what the chunk size is though - I'll add a comment explaining the choice of `4096` as a default.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/763/commits/51e3be7f15a01865d93201b0db18755c87dbf282

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
jonathanhaigh-arm: Rob and I decided that it would be better to embed the version numbers into the image types, so this code has changed: https://github.com/ARMmbed/meta-mbl/pull/763/commits/0953b4823cd59b89736faf359b51927654d3328f

PR #780, opened 2019-11-12
 meta-mbl-apps:mbed-cli: add mercurial-native to DEPENDS

esajaa:
mbed-cli uses mercurial in deploy command, and occasionally with
certain parameters this can lead to a build failure when hg-command
is not present


diego-sueiro:
Can you please change the patch subject to the following?
`meta-mbl-apps:mbed-cli: add mercurial-native to DEPENDS`

meta-mbl-apps/recipes-devtools/mbed-cli/mbed-cli_1.10.0.bb
```diff
@@ -5,6 +5,7 @@
 LICENSE="Apache-2.0"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=4336ad26bb93846e47581adc44c4514d"
 
+DEPENDS = "mercurial-native"
```
jonathanhaigh-arm: I don't think the `-native` bit should be here. mbed-cli depends on `mercurial-native` when it's being built for the host machine, but depends on `mercurial` when it is being built for the target. That's the behaviour we'd get by writing
```
DEPENDS = "mercurial"
```
because when building `-native` recipes, BitBake will automatically convert recipes in `DEPENDS` to be `-native`.

...Shouldn't this be an `RDEPENDS` rather than  a `DEPENDS`? Do we need mercurial to be around when we build mbed-cli or just when it's being run?

```suggestion
RDEPENDS_${PN} += "mercurial"
```

esajaa: Both points make sense. I wasn't aware of the automatic conversion. The RDEPENDS is correct, because mercurial is required runtime. The same applies to git as well, since both are used by the mbed-cli, so in theory it could be added as well.

jonathanhaigh-arm: Thanks for the changes.

jonathanhaigh-arm:
I've created https://jira.arm.com/browse/IOTMBL-2625 for the git dependency problem when building for the target.

jonathanhaigh-arm:
Jenkins build for this branch: http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/190/

jonathanhaigh-arm:
The Jenkins build succeeded. Lava hasn't finished yet, but results are at http://e120856-lin/results.php?image_version=custom&image_number=190&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_device

jonathanhaigh-arm:
FWIW, I think @jjohnson-arm's https://github.com/ARMmbed/meta-mbl/commit/0cd8488e61b84cffc0523013a6eb7d47b82a9952 should fix the issue when building mbl-cloud-client because it now uses `mbed --protocol git` rather than `mbed --protocol ssh`, but if we want the mbed-cli recipe to be useful in its own right rather than just being a tool to build mbl-cloud-client then the change in this PR is still good.

Another note: once the Lava results come back, and if they're clean, then we should still hold off on merging until @diego-sueiro tells us whether to switch the target branch to zeus-dev.

jonathanhaigh-arm:
Lava results: all tests were okay except for rootfs update (which we currently expect to fail) and Wi-Fi access on rpi3b (https://lava.mbedcloudtesting.com/scheduler/job/78297). The last three Wi-Fi access tests on that particular board (https://lava.mbedcloudtesting.com/scheduler/device/bcm2837-rpi-3-b-08) have failed (https://lava.mbedcloudtesting.com/results/78655, https://lava.mbedcloudtesting.com/results/78524, https://lava.mbedcloudtesting.com/results/78442) so it seems to be an issue in the test farm.

jonathanhaigh-arm:
> Another note: once the Lava results come back, and if they're clean, then we should still hold off on merging until @diego-sueiro tells us whether to switch the target branch to zeus-dev.

We've decided to go with warrior-dev for now.

PR #781, opened 2019-11-13
create-update-payload: support creating v1 payloads

jonathanhaigh-arm:
mbl-os-0.9 only supports installing payloads with format version 1, but
mbl-os-0.10 will only support installing payloads with format version 3
(or maybe 4). Therefore, in order to upgrade from mbl-os-0.9 to
mbl-os-0.10, we must be able to create mbl-os-0.10 payloads that have
format version 1 rather than 3.

Give create-update-payload a "--format-version" option that can be used
to specify the payload format version to use, and give the script the
functionality to build version 1 payloads.

For IOTMBL-2613: Robust update: provide mechanism to update across
payload format versions

**Jenkins builds**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/192/ (succeeded; before addressing review comments).
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/195/ (succeeded; after addressing review comments).

**Manual testing done**
* On imx6ul-pico, imx7d-pico, imx8mmevk and rpi3b+:
  * Update using `mbl-firmware-update-manager` and single payload containing all components from warrior-dev -> slightly different warrior-dev.
  * Update using `mbl-firmware-update-manager` from mbl-os-0.9 to warrior-dev. For this update I created a rootfs update with payload version 1 (mbl-os-0.9's format) and a payload containing everything except the rootfs with payload version 3 (warrior-dev's format). I installed the v1 rootfs update onto mbl-os-0.9 then installed the v3 "everything except the rootfs" update. Notes:
    * I had to update the rootfs before the kernel because the warrior-dev kernel component comes with an initramfs that initializes the watchdog, and we need a rootfs that will stroke the watchdog before we can cope with an initramfs that initializes the watchdog.
    * Once the warrior-dev rootfs component is installed, the target device can only deal with v3 payloads, not v1 payloads, so the rest of the components must be installed as part of a v3 payload.
    * mbl-os-0.9 doesn't support multi-component payloads so we can't just install a full warrior-dev component set in one go. 
  * Manual comparison of payloads created using an mbl-os-0.9 version of `create-update-payload` and payloads created using warrior-dev `create-update-payload --format-version 1`.

**Lava results**
* Without the new update tests from @rwalton-arm: http://e120856-lin/results.php?image_version=custom&image_number=195&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_device. There are expected rootfs update failures and some Wi-Fi access failures. I checked out a couple of the Wi-Fi failures and the latest warrior-dev tests on the same boards failed too.
* Results for new update tests from @rwalton-arm:
  * imx6ul: https://lava.mbedcloudtesting.com/results/81094, https://lava.mbedcloudtesting.com/results/81095 (no fails).
  * rpi3: https://lava.mbedcloudtesting.com/results/81097, https://lava.mbedcloudtesting.com/results/81098, https://lava.mbedcloudtesting.com/results/81099, https://lava.mbedcloudtesting.com/results/81100 (no fails).
  * imx8mmevk: https://lava.mbedcloudtesting.com/results/81101, https://lava.mbedcloudtesting.com/results/81102 (one failure which seems due to the Wi-Fi driver kernel stack trace we see during boot).
  * imx7d-pico: https://lava.mbedcloudtesting.com/results/81103, https://lava.mbedcloudtesting.com/results/81104 (no fails).

**Notes**
* I've added an option to specify a payload format version, but it would be useful to be able to specify an mbl-os version and let the script work out which payload format version should be used. I intend to add an option to specify the mbl-os version in a later PR.


meta-mbl-distro/scripts/lib/mbl/update/payloadbuilder.py
```diff
@@ -0,0 +1,173 @@
...
+            bb.fatal(
+                'Unsupported payload format version "{}"'.format(
+                    payload_format_version
+                )
+            )
+
+    @abc.abstractmethod
+    def create_archiver(self):
+        """Create a PayloadArchiver."""
+        ...
```
rwalton-arm: We shouldn't need these three dots. Is this to appease one of the insanity checkers?

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
```diff
@@ -46,12 +42,69 @@ def generate_testinfo(self):
...
+    def __init__(self, app_paths):
+        """
+        Create a TarXzAppsImage object.
+
+        Args:
+        * app_paths list<str>: list of paths to the ipk files to add to the
+          payload.
+
+        """
+        super().__init__(app_paths, image_format_version=3)
```
rwalton-arm: Should we have some kind of global data structure holding the format versions? We could have a single `namedtuple` which contains the format version for each image. I think it's starting to become a bit of a maintenance headache to have to look through all the class definitions and change the versions.

meta-mbl-distro/scripts/lib/mbl/update/payloadbuilder.py
jonathanhaigh-arm: We need _something_ there to satisfy Python itself, either `1`, `pass`, `...` or whatever. Otherwise Python 3.6 says something like: "IndentationError: expected an indented block ".

The [abc module docs](https://docs.python.org/3/library/abc.html) all use `...`, so I did too.

jonathanhaigh-arm: Oh, does the docstring count as the "_something_" I need there?

jonathanhaigh-arm: Yes, it does, cool, I'll remove the `...`s.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/781/commits/3c4543350777adf76dabf3bbe77995ec95162a18

rwalton-arm: yeah the docstring counts as something. I think in python 2.7 you had to have `pass` or something else also, but that changed for python3 IIRC.

meta-mbl-distro/scripts/lib/mbl/update/payload.py
```diff
@@ -9,20 +9,19 @@
...
+
 
 class UpdatePayload:
     """Class for creating update payloads and metadata."""
 
     def __init__(
         self,
+        payload_format_version,
         tinfoil,
         bootloader_components=[],
```
rwalton-arm: I should've caught this in the last review, we shouldn't be using 'mutable default arguments' as they behave especially weirdly in python. The list object is created at parse time, instead of a new list being created on each instantiation of the object, so if it's mutated in any way it will be mutated for all instances of the class. Might not practically make a difference in this case as this class is a singleton anyway, but it's bad practice to do this in general. We should use `None` as the default value here (and for `apps`) instead of an empty list.

jonathanhaigh-arm: Okay, will address in the code. I wonder if I can effectively use tuples to make this go away...

meta-mbl-distro/scripts/lib/mbl/update/appsimage.py
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/781/commits/d0575f20da4a268148336cd94f5dbbb876d52437

meta-mbl-distro/scripts/lib/mbl/update/payload.py
jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/781/commits/200033857e84ab4f64b29f85dcdb417c57d96c35

jonathanhaigh-arm:
@rwalton-arm, can you re-review please?

meta-mbl-distro/scripts/lib/mbl/update/payloadimage.py
```diff
@@ -10,12 +10,47 @@
...
 
+import mbl.update.versionedclassregistry as vcr
 import mbl.util.fileutil as futil
 
 
 class PayloadImage(abc.ABC):
     """ABC for creating image files for components in an update payload."""
 
+    @classmethod
+    def register(cls):
```
rwalton-arm: We might be able to make this cleaner by using `subclasshook` so we don't need to call this register method on every subclass?

meta-mbl-distro/scripts/lib/mbl/update/versionedclassregistry.py
```diff
@@ -0,0 +1,56 @@
...
+    * name str: name of the thing that cls represents.
+    * version int: version of the thing that cls represents.
+    * cls class: class that represents the thing.
+
+    """
+    if name not in _registry:
+        _registry[name] = {}
+
+    assert version not in _registry[name]
+    _registry[name][version] = cls
```
rwalton-arm: I'm wondering if we should use `weakref`s here so we don't prevent garbage collection of these objects?

meta-mbl-distro/scripts/lib/mbl/update/payloadimage.py
jonathanhaigh-arm: I considered doing something like that, but I don't actually want to register every subclass of `PayloadImage`.

meta-mbl-distro/scripts/lib/mbl/update/versionedclassregistry.py
jonathanhaigh-arm: I don't think `weakref`s would help here because the things in the registry are required to create the objects that will have references to them.

We could potentially limit the scope of the registry so that it can be garbage collected once the `PayloadBuilder` has been created, but I don't think its worth it.

meta-mbl-distro/scripts/lib/mbl/update/payloadimage.py
jonathanhaigh-arm: @rwalton-arm and I spoke IRL... I'm going look into making the registration automatic again, even if it doesn't use `subclasshook`.

jonathanhaigh-arm: I can't think of a good way to do this without making things too complicated.

`subclasshook` doesn't really do what we want - it's for customizing which classes are considered subclasses of an ABC. We could potentially use the list of subclasses of the ABC (as defined by `issubclass`) as a registry, but the logic to retrieve things from the registry would get more complicated, and it would require marking some things that _are_ actually subclasses of the ABC as not subclasses as defined by `issubclass`, which is a bit nasty.

Another way to achieve this would be to create a metaclass that automatically registered the correct classes, but the metaclass of the classes involved is already `abc.ABCMeta`, so we'd have to derive a new metaclass from `abc.ABCMeta` and get that to do registration.

I don't think its worth adding this level of complexity in order to remove the calls to the register function, which I've now renamed to `register_versioned_class` from `register` to avoid confusion with the `register` function provided by `abc.ABCMeta`.

rwalton-arm: I'm wondering if we can just maintain a mapping of payload versions to the class objects that will be used in that payload version, rather than keeping a mapping of image version numbers to payload versions. I think we can then basically get rid of the registration mechanism (which seems a bit convoluted to me) and we still have all mappings in one place for easier maintenance.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/781/commits/955f741f2ed28dec2f89a6ae89c04c815fb06f90

PR #795, opened 2019-11-22
initramfs-init-script: don't rely on the rootfs label

jonathanhaigh-arm:
We're going to start using an ext4 image as our rootfs update payload.
At the time the ext4 image is created we have to set the file system's
label, but we don't know whether it will be written to the first or
second rootfs bank.

Therefore, we have to stop relying on the label of the rootfs file
system to determine which bank is active.

Update initramfs-init-script to not rely on the rootfs (or log)
partition labels, but use the partition numbers that are left in
/config/factory/part-info instead.

I've added a new MBL_PART_INFO_DIR variable to use in the initramfs
script, so be consistent and use it in the base-files bbappend too.

**Jenkins builds**
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/200/ (before changes for review)
* http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/201/ (after changes for review)

**Lava links**
* http://e120856-lin/results.php?image_version=custom&image_number=201&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_device. Failures:
  * https://lava.mbedcloudtesting.com/scheduler/job/83641 (Wi-Fi failure; common).
  * https://lava.mbedcloudtesting.com/scheduler/job/83652 (Wi-Fi failure; common).
  * https://lava.mbedcloudtesting.com/scheduler/job/83628 (lxc-create-action timed out after 600 seconds; common).
  * https://lava.mbedcloudtesting.com/scheduler/job/83650 (optee_test failure... optee_test exited `255` during regression test 6016 ("Storage concurency" \[sic\]). It doesn't seem like a normal optee_test failure - I don't think the test finished and returned a failure code, I think it just crashed. I don't see how this failure could be anything to do with my changes, but I re-submitted the lava job to see what happens... It passed (https://lava.mbedcloudtesting.com/scheduler/job/85451). I've created https://jira.arm.com/browse/IOTMBL-2659.

PR #797, opened 2019-11-25
partitions: Add variable for device on which MBL lives

jonathanhaigh-arm:
It will be useful for our rootfs installer to know the name of the
storage device on which the MBL partitions live, so add a variables for
the devices on which each partition resides.

In particular, this is useful because we will no longer be able to rely
on finding rootfs partitions by their labels.

The values of these variables will be written to the "part-info"
directory on devices. Currently all the MBL partitions must be on the
same device but this will be useful in the future if we start to support
using multiple storage devices.

For IOTMBL-2656: Robust update: don't rely on root filesystem labels

**Jenkins builds**
* http://jenkins.mbed-linux.arm.com/job/jh-test/204/ (cancelled)
* http://jenkins.mbed-linux.arm.com/job/jh-test/205/ (in progress; after variable name change)

**Lava results**
* http://e120856-lin/results.php?image_version=custom&image_number=205&custom_version=jh-test&lava_query=&lava_owner=&failures=false&submitter=jonhai01&sort=by_device. Failures:
  * https://lava.mbedcloudtesting.com/scheduler/job/85660 (Wi-Fi access; common failure).
  * https://lava.mbedcloudtesting.com/scheduler/job/85663 (Wi-Fi access; common failure).
  * https://lava.mbedcloudtesting.com/scheduler/job/85649 (lxc-create-action timeout; common failure).
  * https://lava.mbedcloudtesting.com/scheduler/job/85647 (lxc-create-action timeout; common failure).
  * https://lava.mbedcloudtesting.com/scheduler/job/85661 (optee_test failure, similar to https://github.com/ARMmbed/meta-mbl/pull/795 / https://jira.arm.com/browse/IOTMBL-2659).
  * https://lava.mbedcloudtesting.com/scheduler/job/85659 (rootfs image update via pelion; SSH failure after update, https://jira.arm.com/browse/IOTMBL-2660).

**Manual testing**
* Ran each image from artifactory for jh-test job 205 and checked that the `_DEVICE_NAME_` files in /config/factory/part-info look correct.


meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -49,6 +49,10 @@ MBL_DEFAULT_FS_FREQ ?= "0"
 # force the fsck to run.
 MBL_DEFAULT_FS_PASSNO ?= "2"
 
+# Default device name (file name in /dev) for the storage device on which
+# partitions live.
+MBL_DEFAULT_DEVICE_NAME ?= "${MBL_WKS_STORAGE_DEVICE_NAME}"
```
diego-sueiro: Shouldn't we add `STORAGE` word in this variable name?

jonathanhaigh-arm: No...but yes.

The `MBL_DEFAULT_` variables in this file all follow the naming convention `MBL_DEFAULT_` + suffix of partition specific variables for which this is the default. E.g. `MBL_DEFAULT_FS_PASSNO` is the default value for `MBL_FACTORY_CONFIG_FS_PASSNO`, `MBL_HOME_FS_PASSNO`, etc. The `_process_var_with_default` function relies on this convention (which I admit isn't very nice).

The partition specific variables don't really want `STORAGE` in their names - it's obvious that they relate to storage because they relate to partitions. `_DEVICE_NAME` seems like a reasonable suffix for the partition specific variables...

Having said that, the `MBL_DEFAULT_` prefix convention for these variables is a bit rubbish, especially since the names make their way into every recipe. `MBL_DEFAULT_DEVICE_NAME` is a particularly bad example.

How about if we change the `MBL_DEFAULT_` prefix for all of these variables to `MBL_PARTITIONS_DEFAULT_`?  So we'd end up with `MBL_PARTITIONS_DEFAULT_DEVICE_NAME`, `MBL_PARTITIONS_DEFAULT_FS_PASSNO` etc.

diego-sueiro: I much prefer the `MBL_PARTITIONS_DEFAULT_` prefix. It makes more clear the usage of the this set of variables.
If you decide to go into this path, we need to check if the user documentation needs to be updated as well.

jonathanhaigh-arm: The user docs don't mention the `MBL_DEFAULT` variables.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/797/commits/c06c592c16b4be984ed41d2e307a6a95124e201d

jonathanhaigh-arm:
@rwalton-arm, what do you think of https://lava.mbedcloudtesting.com/scheduler/job/85659? Does the test need to check that networking is up properly before doing the /etc/build mtime test?

PR #811, opened 2019-12-02
swupdate: Add swupdate-handlers recipe

rwalton-arm:
This commit adds a recipe that compiles our swupdate handlers to
object files, which need to be linked with the swupdate handlers
kbuild module. The do_install task of this recipe moves the object
files to a staging directory, where they can be picked up by our
swupdate bbappend and linked with the swupdate handlers module.

We also patch swupdate to tell it to link our handlers and add a
menuconfig item to disable and enable the handlers if needed (the
default is to have our handlers enabled).

Also remove any reference to rootfs_installer.sh from the mbl-cloud-client
recipe.

JIRA: IOTMBL-2586
Jenkins: http://jenkins.mbed-linux.arm.com/job/rw-test/209/

Related PRs: 
* https://github.com/ARMmbed/mbl-core/pull/261
* https://github.com/ARMmbed/mbl-core/pull/270

NOTE to self: SRCREV must be updated before merging


meta-mbl-apps/recipes-core/swupdate-handlers/swupdate-handlers.bb
```diff
@@ -0,0 +1,43 @@
...
+-DBOOTFLAGS_DIR=${MBL_BOOTFLAGS_DIR} \
+-DUPDATE_PAYLOAD_DIR=${MBL_SCRATCH_DIR} \
+-DLOG_DIR=${localstatedir}/log \
+-DROOTFS_TYPE=${MBL_ROOT_FSTYPE} \
+-DFACTORY_CONFIG_PARTITION=${MBL_FACTORY_CONFIG_DIR} \
+"
+
+# We fetch the swupdate headers and shove them into our handler's include dir, as our handlers depend on some
+# of swupdate's headers.
+do_fetch_headers() {
```
diego-sueiro: I don't think this is the right approach.

Why we can't add `DEPENDS = "swupdate"`? This way bitbake will add all headers and generated binaries in the `${WORKDIR}/recipe-sysroot`.

rwalton-arm: We need inter-task dependencies because:
* We compile swupdate-handlers to object files which need access to swupdate's headers
* We then move the swupdate-handlers object files to the swupdate sourcetree before swupdate's `do_compile` task runs, so the handlers can link with swupdate's "handler" kbuild module
* If we use DEPENDS we can't shove our handlers back into the swupdate sourcetree before its `do_compile` task runs IIUC

jonathanhaigh-arm: I wonder if a cleaner way to do this would be to have a new `swupdate-headers` recipe that just installs headers into somewhere like `${includedir}/swupdate`. Then the `swupdate-handlers` recipe could just `DEPEND` on `swupdate-headers` and use `do_install` or `do_deploy` to make the MBL handlers available to the swupdate recipe.

I haven't read all of the recipe changes yet, but bypassing `do_deploy` and `do_install` to share things between recipes is hard to get right.

diego-sueiro: @rwalton-arm, 

Thanks for the explanation.

I think this is not the right way of doing this. In this case, we should have everything concentrated in the swupdate bbappend.

jonathanhaigh-arm: There are benefits to building our handlers outside of swupdate's build system:
* We can use extra warning flags when we compile.
* We can set up static analysis of our handlers.
* We can split code into multiple files without battling with swupdate's build system.
* We can potentially write handlers in C++ rather than C.

diego-sueiro: We can still accomplish most of these tasks by having separate tasks functions.

This interdependency of the tasks between different recipes and files copy is not easy to maintain and looks weird.

rwalton-arm: I'm not sure if littering the swupdate recipe with extra tasks is much better/clearer than having a separate recipe with inter task dependencies. Also, I don't particularly want to add extra compile flags to swupdate's build system as I think that's probably a can of worms we don't need to open at this point.

I would prefer it if I didn't have to copy the files to the deploy directory, and was hoping someone would have a suggestion which got me the same behaviour as `DEPENDS` (headers automatically available in the sysroot) using inter task dependencies. One possibility is to add our handlers as patches to swupdate, and use swupdate's build system as is, but that loses us the benefits Jonathan mentioned above. Another thing, in the future I thought we wanted to use swupdate more as a shared library as an `RDEPENDS` of UpdateD, so we can reuse stuff out of the swupdate core & write the more complex handlers in C++?

I'm not yet convinced that any of the suggestions in this thread are better than what I've got at the moment. Maybe just adding patches is the best approach, but then the next comment is going to be "why aren't we upstreaming the patches?" and I don't think our handlers are generic enough to be upstreamed at this point. I'll think about it and decide what I'm going to do in the morning...

diego-sueiro: If we decide to keep it as a separate recipe, the @jonathanhaigh-arm suggestion of a `swupdate-headers` recipe is a good idea. And the `swupdate-handlers` recipe generates the object files needed by the main `swupdate` recipe. The main `swupdate` recipe then needs to `DEPENDS` on `swupdate-handlers` and at the do_configure or do_compile task you can copy the object files from the `recipe-sysroot` to the `handlers` folder.

diego-sueiro: Just adding some ideas on how to set things in the `swupdate-handlers` recipe if we decide to follow what I suggested above.

You can have the following:
```
DEPENDS += "swupdate-headers"
do_install() {
    install -d ${D}/${libdir}/
    install -m 666 ${B}/CMakeFiles/swupdate-handlers.dir/*.c.o ${libdir}/
}
FILES_${PN} = ""
FILES_${PN}-dev += "${libdir}/*.c.o"
```

jonathanhaigh-arm: One thing that Rob and I discussed yesterday is: if we have a `swupdate-headers` recipe, how do we keep the `SRCREV` consistent between the `swupdate` and `swupdate-headers` recipes?

Another thing that I've been thinking about is: in order to reduce the fragility of the recipes, we'll want to assume as little as possible about what the `swupdate` recipe actually does. One thing that concerns me is that we assume that swupdate's `do_configure` doesn't modify the header files that we use. We also assume things about where object files must be inserted into the swupdate build.

This might be controversial, but what about something like:
1. We have a swupdate-headers recipe that:
    * `requires` the swupdate recipe, ensuring that we're using the same `SRCREV` as swupdate.
    * Use the swupdate recipe's `fetch`, `patch` and `configure` tasks to make sure we're using properly patched and configured headers.
    * Has an empty `build` task.
    * Installs the headers into a standard place using an overridden `install` task, allowing the swupdate-handlers recipe to just `DEPEND` on `swupdate-headers` and then access them automatically using `#include <swupdate/blah.h>`.

1. In the swupdate-handlers code in mbl-core, we:
    * Create a proper library rather than just an object file.
    * Use CMake to install the the library into a standard location.
    This allows the swupdate recipe to just `DEPEND` on the swupdate-handlers recipe and link the handlers in a standard way.

1. In the swupdate bbappend, all we do is:
    * `DEPEND` on `swupdate-handlers`.
    * Arrange for `-lswupdate-handlers` to be passed to the linker.
    * Patch `handlers.c` (or whatever it's called) to register thin wrappers for our handlers. I.e.
      ```C
      int rootfsv4_handler_wrapper(struct img_type *img, void *data)
      {
          return rootfsv4_handler(img, data);
      }
      __attribute__((constructor))
      void rootfsv4_handler_init(void)
      {
          register_handler("ROOTFSv4"
                           , rootfsv4_handler
                           , IMAGE_HANDLER
                           , NULL);
      }
      ```

Just adding our handlers as `C` files into the swupdate recipe is temptingly simple, but having extra warnings from the compiler due to us building the rootfs handler separately has already caught bugs, so I'd be reluctant to give that up. I'm not really convinced one way or the other yet.

diego-sueiro: @jonathanhaigh-arm,

Your proposal makes sense to me.

rwalton-arm: Jonathan and I discussed this proposal earlier and I'm also on board.

meta-mbl-distro/classes/mbl-partitions.bbclass
```diff
@@ -219,7 +219,6 @@ MBL_PARTITION_NAMES += "BOOT"
 # ROOT (file system; two banks)
 # ------------------------------------------------------------------------------
 MBL_ROOT_MOUNT_POINT = "/"
-MBL_ROOT_LABEL ?= "rootfs"
```
diego-sueiro: I'm not comfortable with getting rid of the partition label. I do understand that by copying the image file, the label of the partition is get overridden and we need to rely on the partition number store in the `part-info`, instead.

I'm just thinking about user applications for whatever reason will rely on the partition label (or even tests to validate the update). Because we know what partition number is the bank we are flashing to, we could restore the label after the copy.

jonathanhaigh-arm: > we could restore the label after the copy.

Not if we start using e.g. squashfs.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
```diff
@@ -0,0 +1,21 @@
...
+require swupdate-append-common.inc
+
+SUMMARY = "Header files for building handlers for swupdate"
+DESCRIPTION = "${SUMMARY}"
+FILESEXTRAPATHS_prepend = "${THISDIR}/../../../../meta-swupdate/recipes-support/swupdate/swupdate:"
+
+PACKAGES = "${PN}"
+FILES_${PN} = "${includedir}/*" 
+
+do_compile() {
```
diego-sueiro:
```suggestion
do_compile[noexec] = "1"
```

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: MIT
+
+require recipes-support/swupdate/swupdate_${PV}.bb
+require swupdate-append-common.inc
+
+SUMMARY = "Header files for building handlers for swupdate"
+DESCRIPTION = "${SUMMARY}"
+FILESEXTRAPATHS_prepend = "${THISDIR}/../../../../meta-swupdate/recipes-support/swupdate/swupdate:"
```
diego-sueiro: I'm not entirely sure if this is needed. I'm expecting that the `require` will bring the `FILESPATH` from the required recipe.
If not, can you try to use one of these two options instead of "../'s"?
`FILESEXTRAPATHS_prepend = "meta-swupdate/recipes-support/swupdate/swupdate:"`
or
`FILESEXTRAPATHS_prepend = "${OEROOT}/layers/meta-swupdate/recipes-support/swupdate/swupdate:"`

jonathanhaigh-arm: > I'm expecting that the `require` will bring the `FILESPATH` from the required recipe.

It doesn't, unfortunately.

 > `FILESEXTRAPATHS_prepend = "meta-swupdate/recipes-support/swupdate/swupdate:"`

I tried that, but `FILESPATH` entries are absolute paths.

> `FILESEXTRAPATHS_prepend = "${OEROOT}/layers/meta-swupdate/recipes-support/swupdate/swupdate:"`

I thought I'd tried that, but I think maybe I had a typo in the path and thought it wouldn't work when in fact, it does... Or at least `bitbake -e swupdate-headers` tells me that `$OEROOT` has a reasonable value. 

Thanks for forcing me to try it again :-)

TBH, I still don't like this line even when using `${OEROOT}` - it still makes an assumption about where we keep our layers. I can't think of anything better though.

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/811/commits/231556c65ee51a13283f25ca8f8c658d83379bb0

meta-mbl-distro/conf/distro/include/mbl-distro.inc
```diff
@@ -31,6 +31,11 @@ BUILDHISTORY_COMMIT = "1"
 
 IMAGE_FSTYPES = "ext4 wic.gz wic.bmap tar.xz"
 
+# Create minimal rootfs file system images
+IMAGE_ROOTFS_SIZE = "0"
+IMAGE_OVERHEAD_FACTOR = "1"
```
diego-sueiro: By reading the [Yocto manual](https://www.yoctoproject.org/docs/current/mega-manual/mega-manual.html#var-IMAGE_OVERHEAD_FACTOR) we need to have some extra space for the post install scripts processing. I'd rather use at least 1.1 factor.

diego-sueiro: But at the same time the manual is not clear if this overhead factor is only applied when the whole rootfs processing was finished.

diego-sueiro: Looking at the `image.bbclass` it seems that the `IMAGE_OVERHEAD_FACTOR` is taking into account after the rootfs has being generated. So, I think that in this case, it is safe to use the factor 1.

jonathanhaigh-arm: Thanks for checking; I'm still a bit unsure about this.

IIUC the rootfs size we'll get given the `IMAGE_ROOTFS_SIZE` and `IMAGE_OVERHEAD_FACTOR` that I set is still determined using the output of `du` for the sysroot, which is a bit dodgy, especially if we don't add a bit extra to make up for any differences between what `du` thinks and what is actually required.

Unfortunately, I don't think there's a way to create an ext4 image that's exactly the right size for a bunch of files - you've got to specify the fs size when creating the image before copying the data into it.

I'm not too worried about this because:
* If the rootfs size is too small we should get an error at build time. If there's a problem at runtime then it means were writing to the rootfs at runtime, which we shouldn't be doing.
* Hopefully once we switch to squashfs BitBake will provide a mechanism to do exactly what we want without worrying about these numbers.

diego-sueiro: Another thing that came up in my mind. Should we have, for some reason, an image size that is multiple of flash erase size?

jonathanhaigh-arm: I don't *think* that would buy us anything.

Something that might be useful is tuning the sizes that the file system uses for various types of "chunk" based on properties of the flash storage. I don't think it's worth bothering with that before we switch to squashfs though.

jonathanhaigh-arm: I think tuning the file system's "chunk" sizes would be much more useful for the r/w partitions TBH - after all, the flash erase size is only a factor when doing writes.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
diego-sueiro: > > I'm expecting that the `require` will bring the `FILESPATH` from the required recipe.
> 
> It doesn't, unfortunately.
> 
Maybe this is only valid when you directly require/include a `.inc` file? Can you please try that?
`require recipes-support/swupdate/swupdate.inc` and remove the `FILESEXTRAPATHS_prepend` and see how it goes?

> TBH, I still don't like this line even when using `${OEROOT}` - it still makes an assumption about where we keep our layers. I can't think of anything better though.
Me neither. That's why I'm trying to explore other possibilities, but I'm not sure if you already tried what I'm proposing.

jonathanhaigh-arm: > Maybe this is only valid when you directly require/include a .inc file? Can you please try that?

Even if `require`ing swupdate.inc without using an absolute path would work, I don't think that would be a good solution because the `SRCREV` comes from the swupdate `bb` file, not the `inc` file, and obtaining the correct `SRCREV` was one of the main motivations for this approach.

meta-mbl-distro/scripts/lib/mbl/update/payloadbuilder.py
```diff
@@ -35,6 +35,14 @@ class PayloadBuilder:
             ri.MBL_ROOTFS_ID: ri.RootfsImageV3,
             ai.MBL_APPS_ID: ai.AppsImageV3,
         },
+        4: {
+            pa.MBL_PAYLOAD_ARCHIVER_ID: pa.PayloadArchiverV3,
+            wi.MBL_WKS_BOOTLOADER1_ID: wi.WksBootloader1ImageV3,
+            wi.MBL_WKS_BOOTLOADER2_ID: wi.WksBootloader2ImageV3,
+            bi.MBL_BOOT_ID: bi.BootImageV3,
+            ri.MBL_ROOTFS_ID: ri.RootfsImageV4,
```
diego-sueiro: Are we introducing any backward incompatibility with this new version?

jonathanhaigh-arm: Depends what you mean. The new version of the update scripts won't be able to install old versions of payloads, but the payload script will still be able to create payloads using the old format version so you'll be able to update from 0.10 to 0.11 with a payload created in a 0.11 workarea by using `--mbl-os-version 0.10`.

If what you're thinking is "Do we need to update the *Version specific upgrade notes* section in the docs?" then the answer is "Yes".

jonathanhaigh-arm: I suppose another question is "Should we support updating directly from 0.9 to 0.11?". Technically, we might be able to make that work using the same flow as 0.9 -> 0.10 updates, but TBH, I'm not convinced it's worth it. I suggest that except for 0.9 -> 0.10, we don't support upgrading from 0.9 because 0.9 doesn't support multi-component updates. Going from e.g. 0.10 -> 0.12 directly should be simpler.

diego-sueiro: > I suppose another question is "Should we support updating directly from 0.9 to 0.11?". Technically, we might be able to make that work using the same flow as 0.9 -> 0.10 updates, but TBH, I'm not convinced it's worth it. I suggest that except for 0.9 -> 0.10, we don't support upgrading from 0.9 because 0.9 doesn't support multi-component updates. Going from e.g. 0.10 -> 0.12 directly should be simpler.

I don't think we should support 0.9 -> 0.11, because if we do then we will have to do from all future versions.
As soon as we make clear the _special_ step about updating from 0.9, I'm happy with it.

jonathanhaigh-arm: Okay. There are also other related docs edits I need to do due to the rootfs label changes (including changes to the create-partition-table-markdown script).

diego-sueiro: And the part of how to check if the rootfs was updated.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
```diff
@@ -0,0 +1,20 @@
...
+
+require recipes-support/swupdate/swupdate_${PV}.bb
+require swupdate-append-common.inc
+
+SUMMARY = "Header files for building handlers for swupdate"
+DESCRIPTION = "${SUMMARY}"
+FILESEXTRAPATHS_prepend = "${OEROOT}/layers/meta-swupdate/recipes-support/swupdate/swupdate:"
+
+PACKAGES = "${PN}"
+FILES_${PN} = "${includedir}/*"
```
diego-sueiro: The header files should be included in the `-dev` package, not in the main one.

meta-mbl-apps/recipes-support/swupdate/files/arm-handlers.c
```diff
@@ -0,0 +1,19 @@
...
+
+int rootfsv4_handler_wrapper(struct img_type *img, void *data)
+{
+    return rootfsv4_handler(img, data);
+}
+
+__attribute__((constructor))
+void rootfs_handler_init(void)
+{
+    register_handler("ROOTFSv4"
```
diego-sueiro: Can't we use a `#define` (and potentially some macros) that can be passed at compile time to determine what version we are going to use?

I'm just worried about the maintenance effort when we decide to support multiple versions.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
jonathanhaigh-arm: I've been debating something like that with myself for a few days...

If I didn't know anything about the recipes, I'd expect to find headers in "swupdate-dev", but  we don't want to use that name to avoid clashes with the swupdate recipe. In my mind the name "swupdate-headers" means "unofficial-swupdate-dev", or "not-quite-swupdate-dev"... I.e. it's sort-of an unofficial part of the swupdate recipe rather than a completely new thing that should have it's own dev package ("swupdate-headers-dev").

I'm not convinced either way though.

meta-mbl-apps/recipes-support/swupdate/files/arm-handlers.c
jonathanhaigh-arm: I started to write a response to this, but was starting to get really long. I think it would be easier to speak in person. @diego-sueiro I'll try to catch you after lunch.

jonathanhaigh-arm: @diego-sueiro and I have spoken about this.

I think the short answer is: not without either:
* Adding another place that knows about image format versions.
* Making things even more complicated.

rwalton-arm: I think there's at least couple of options:
* Register all our existing handlers and use the sw-description to determine which version to use
* Have an ifdef here which selects the handlers to register based on the payload version which is set in a config file somewhere

I'm not convinced it's worth adding ifdefs when swupdate will select the handler automatically based on the sw-description.

rwalton-arm: > 
> 
> @diego-sueiro and I have spoken about this.
> 
> I think the short answer is: not without either:
> 
>     * Adding another place that knows about image format versions.
> 
>     * Making things even more complicated.

Missed this when I was typing my previous response. Agreed.

jonathanhaigh-arm: Sorry Rob, saying that Diego and I have spoken about this isn't very useful information for you!

> Register all our existing handlers and use the sw-description to determine which version to use

I don't think that would alleviate the issue that @diego-sueiro has identified - unless we do something else as well, we'd still need to mention all the versions of the handlers in this file. Maybe I'm misunderstanding what you mean...

> Have an ifdef here which selects the handlers to register based on the payload version which is set in a config file somewhere

Right, yeah this is sort-of what I was getting at:
* We could add some config to our BitBake environment. That would add another place that knows about image format versions. We could use it to remove knowledge from this file (arm-handlers.c), but both `create-update-payload` and the swupdate-handlers code would still need to know about all the versions of handlers. We'd still have three places that need updating when we add new handlers.

  We could potentially use config in the BitBake environment to cause build time errors if not all three of `swupdate-handlers`, `swupdate` and `create-update-payload` agree on what the latest versions are, but we've got reasonable automated tests that should catch these issues already anyway. And in some cases we'd get a link error for swupdate if we updated swupdate-handlers and forgot to update here.

* We could try to extract the version information that arm-handlers.c needs from swupdate-handlers or `create-update-payload`, but I think that would get complicated.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
diego-sueiro: By default bitbake put the headers in the `-dev` package and when you have `DEPENDS` on that package, bitbake install the `-dev` package in the `recipe-sysroot` directory. This is the way it works.

By adding headers in the main package we are deviating from the standard. If we really follow this path we should at least add a comment in the recipe explaning the reason for not following the standard.

meta-mbl-apps/recipes-support/swupdate/files/arm-handlers.c
diego-sueiro: I'm just worried about maintenance and potential backward incompatibilities.

For the maintenance, when we decide to up-rev any payload version, what source code should we modify?
If this is well documented or even generate build-time errors, I don't see any issues.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
jonathanhaigh-arm: > we should at least add a comment in the recipe explaning the reason for not following the standard.

Good point. There are a lot of non-obvious things going on in the recipes we've added and modified here. I'll add some comments.

meta-mbl-apps/recipes-support/swupdate/files/arm-handlers.c
jonathanhaigh-arm: I'll add some comments where there's non-obvious coupling between code.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
jonathanhaigh-arm: I've added some comments in https://github.com/ARMmbed/meta-mbl/pull/811/commits/e328c7cde35a9fdd84e61ef81e7db49508a3490e.

meta-mbl-apps/recipes-support/swupdate/files/arm-handlers.c
jonathanhaigh-arm: I've added some comments. Some are in https://github.com/ARMmbed/meta-mbl/pull/811/commits/e328c7cde35a9fdd84e61ef81e7db49508a3490e, most are in the swupdate-handlers code in https://github.com/ARMmbed/mbl-core/pull/261/commits/ac4df7db4161bfe6e1708c047aab4ab8b7b04aee and https://github.com/ARMmbed/mbl-core/pull/261/commits/9305879efe61138ae4d536dd29fb599f4b7e1ff1.

jonathanhaigh-arm:
@rwalton-arm, I made some changes to the CMake config in 9b9a0b34f97cc86e5495b7ede9035e10f48e1c40. Can you review please?

rwalton-arm:
> @rwalton-arm, I made some changes to the CMake config in [9b9a0b3](https://github.com/ARMmbed/meta-mbl/commit/9b9a0b34f97cc86e5495b7ede9035e10f48e1c40). Can you review please?

The CMake changes look fine to me.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
diego-sueiro: Thanks for adding the comments, but you still haven't convinced me why we are not following the yocto standard and installing headers in the `swupdate-headers` package instead of `swupdate-headers-dev`.

For instance, on the oe-core we have the `swupdate-headers` and the headers are installed in the `swupdate-headers-dev` package.

jonathanhaigh-arm: > For instance, on the oe-core we have the swupdate-headers and the headers are installed in the swupdate-headers-dev package.

I don't know what you mean about oe-core... swupdate isn't in oe-core.

Anyway, https://github.com/ARMmbed/meta-mbl/pull/811/commits/e85d5b822ac6811c23929b349f11bb972f2a835a.

diego-sueiro: Sorry. There was a typo in my comment. For the oe-core I meant `linux-libc-headers` not `swupdate-headers`.

meta-mbl-apps/recipes-support/swupdate/swupdate-headers_2019.11.bb
```diff
@@ -12,9 +12,7 @@ SUMMARY = "Header files for building handlers for swupdate"
 DESCRIPTION = "${SUMMARY}"
 FILESEXTRAPATHS_prepend = "${OEROOT}/layers/meta-swupdate/recipes-support/swupdate/swupdate:"
 
-PACKAGES = "${PN}-dev"
-FILES_${PN}-dev = "${includedir}/*"
```
jonathanhaigh-arm: I don't think we should remove these lines: `swupdate_${PV}.bb` adds a bunch of extra packages that we don't want, and might set `FILES_${PN}-dev` to something other than what we want here.

I wonder if the error you saw was because the default value of `RDEPENDS_${PN}-dev` contains `${PN}`... I suggest that instead of deleting these lines and adding `ALLOW_EMPTY_${PN}`, we try setting `RDEPENDS_${PN}-dev = ""`

I'll try that out...

jonathanhaigh-arm: https://github.com/ARMmbed/meta-mbl/pull/811/commits/010e1522473e8bf622e86befc9ef3004a157a875

jonathanhaigh-arm:
Jenkins build with latest code from here and from https://github.com/ARMmbed/mbl-core/pull/261: http://jenkins.mbed-linux.arm.com/view/jh/job/jh-test/216/

PR #812, opened 2019-12-02
meta-mbl-apps: Add updated recipe and systemd service file

rwalton-arm:
This commit adds a recipe and systemd service file for our update
daemon.

JIRA: IOTMBL-2649
Jenkins build coming soon
Related PRs: https://github.com/ARMmbed/mbl-core/pull/262 (merge after)
Note to self: update srcrev before merging


meta-mbl-apps/recipes-core/updated/files/updated.service
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2019 Arm Limited and Contributors. All rights reserved.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
jonathanhaigh-arm: I'm skeptical of the implicit `Type=simple` here: IIUC it means we don't have a way to let systemd know about initialization failures.

From what I've read, I think we have three real options:
1. Use `Type=forking`, then in the updated code, do initialization and then fork.
2. Use `Type=notify` and have a mechanism to notify systemd that initialization has succeeded.
3. Use `Type=dbus` and commit to using dbus for UpdateD's IPC. Does systemd only support using the system bus for this though? Would we be happy to give up on our MBL/Cloud bus?

If we plan to use `Type=forking` in the future, then I think we should start using it now and add code to do the fork in UpdateD now.

If we plan to use `Type=notify` or `Type=dbus` in the future then using `Type=simple` now and then switching later would probably be reasonable. If we choose `Type=dbus` then using `Type=simple` now and then switching later is probably the only reasonable thing to do because currently our process won't be putting anything on a dbus bus.

I haven't yet looked into how a process can go about notifying systemd that its initialization has finished when using `Type=notify`. IPC rarely turns out to be simple though, so I don't think we should choose `Type=notify` without careful thought.

With my current knowledge, `forking` seems to be the safest option, but my knowledge of the other two `Type`s is very limited.

@diego-sueiro, what do you think?

diego-sueiro: With the integration with Pelion Edge, all cloud connectivity will be handled by them. With this in mind, I suppose we will not be using the mbl-cloud-client/mbl-dbus-cloud anymore. Or if we do , it will be as a replacement of Pelion Edged for cases where the production is a high end linux node and UpdateD should work independently of the cloud connectivity implementation.
We can't guarantee at the moment that Pelion Edged will be using D-bus as an IPC mechanism and a way of getting Systemd to manage/monitor its status.

Since the UpdateD is a daemon it should be started as such (by calling fork after initialization) and the service file should use `Type=forking`.

A proper signal handling needs to be implemented as well, but maybe not at this stage.

jonathanhaigh-arm: > A proper signal handling needs to be implemented as well, but maybe not at this stage.

Yep, we have https://jira.arm.com/browse/IOTMBL-2651 for that.

rwalton-arm: I read some conflicting information online around `Type=forking`. The [systemd manual](https://www.freedesktop.org/software/systemd/man/daemon.html#Writing%20systemd%20Unit%20Files) states "If possible, do not use the Type=forking setting in service files. But if you do, make sure to set the PID file path using PIDFile=. See systemd.service(5) for details." but doesn't provide any rationale for that AFAICT.

jonathanhaigh-arm: > But if you do, make sure to set the PID file path using PIDFile=

Ah, yeah, good point.

I had a quick look at how to notify that the system is initialized when using `Type=notify` (using `sd_notify`), it looks quite simple, and we wouldn't have to manually maintain a PID file, so I'm warming up to it. And if we used `Type=forking` then I think we *would* have to maintain a PID file because the current plan is that UpdateD will fork swupdate processes so systemd wouldn't be able to reliably determine the PID of our main process without our help.

If we use systemd for watchdogging of UpdateD in the future (which I think is our current plan) then we'll probably want to use `sd_notify` from UpdateD to tickle the watchdog anyway.

If we were to use `Type=notify`, though, it might make it a lot harder to support an initialize then fork model in the future. If we don't use the initialize then fork scheme from the beginning then we could end up mixing initialization code and code that would normally end up in the child process together in such a way that it wouldn't be trivial to separate it later.

I'm not convinced that that's a good enough reason to dismiss the simplicity of the `Type=notify` solution though. I think I'm more convinced that `Type=notify` is the way to go now that @rwalton-arm has pointed out the PID file thing.

rwalton-arm: I don't think we should be putting any forking code in our daemon; we should be letting systemd handle all of that. IIUC we should be creating a "new style daemon" (using `Type=notify` or `Type=simple`) unless we have to support a SysVinit system. Reference : https://www.freedesktop.org/software/systemd/man/daemon.html#New-Style%20Daemons

diego-sueiro: > I don't think we should be putting any forking code in our daemon; we should be letting systemd handle all of that. IIUC we should be creating a "new style daemon" (using `Type=notify` or `Type=simple`) unless we have to support a SysVinit system. Reference : https://www.freedesktop.org/software/systemd/man/daemon.html#New-Style%20Daemons

Thanks for the reference. This is a very interesting material.
I'm more than happy if we follow the recommended approach.

jonathanhaigh-arm: Okay, now that we've agreed on the approach to take I don't mind whether we switch to using `Type=notify` now and add notification code to UpdateD or whether we leave the implicit `Type=simple` and do the notification code later. If we leave it until later then we should add a story for it though.

rwalton-arm: I'm leaning toward leaving the implicit `Type=simple` for now, I'll add a task to add the notification code later.

rwalton-arm: I've created https://jira.arm.com/browse/IOTMBL-2672 to track the `sd_notify` work.

PR #827, opened 2019-12-16
meta-mbl-apps:UpdateD: Enable systemd notifications.

rwalton-arm:
We want UpdateD to be able to notify systemd after initialisation, to
inform systemd whether it started successfully. We potentially also want
to use the sd_notify mechanism to stroke the watchdog while UpdateD is
running, however, this is not part of this commit.

This patch will:

* Set the service type to 'notify' in updated.service
* Add a DEPENDS systemd to the UpdateD recipe, so it can use
libsystemd's headers

JIRA: IOTMBL-2672

Related PRs:
* https://github.com/ARMmbed/mbl-core/pull/274


rwalton-arm:
@jonathanhaigh-arm I've updated the mbl-core SRCREV, could you reapprove please?

PR #831, opened 2019-12-20
meta-freescale-mbl:imx-boot:imx8mmevk-mbl: Update and unify patch

diego-sueiro:
The patch to modify the imx-boot image generation to just include the
ATF has to be updated to remove the recently added DEK blob entry.
We took this opportunity to remove all unneeded entries (u-boot, optee,
fdt and dek) in a single unified patch.


diego-sueiro:
Jenkins:
http://jenkins.mbed-linux.arm.com/job/ds-master/12/

PR #833, opened 2020-01-17
updated: Add spdlog as a dependency

rwalton-arm:
This patch adds spdlog to UpdateD's DEPENDS.

Also add a configuration variable to launch UpdateD in 'development'
or 'release' mode. For now this just influences the log level. We
use the INFO level for non-release mode and ERROR in release mode.

*Jenkins*
http://jenkins.mbed-linux.arm.com/job/rw-test/218/


meta-mbl-apps/recipes-support/spdlog/spdlog_%.bbappend
```diff
@@ -0,0 +1 @@
+SRCREV = "1549ff12f1aa61ffc4d9a8727c519034724392a0"
```
diego-sueiro: Can you please add a comment clarifying the reason for picking this specific SHA?

rwalton-arm: Added it when I fixed the licensing

